PR #39, opened 2025-04-06
Cocoapod swift analytics to avoid modular headers

salami:
Switch to swift-based analytics library via Cocoapod


salami:
ignore this... it was pulling in from the fork

PR #55, opened 2025-06-02
Analytics updates

salami:
https://linear.app/tryhelium/issue/HEL-964/analytics-updates

### DIFF
--- a/Sources/Helium/HeliumUIKitViewController.swift
+++ b/Sources/Helium/HeliumUIKitViewController.swift
@@ -16,9 +16,22 @@ public class HeliumPaywallPresentationState: ObservableObject {
     private let useAppearanceToSetIsOpen: Bool
     init(useAppearanceToSetIsOpen: Bool = false) {
         self.useAppearanceToSetIsOpen = useAppearanceToSetIsOpen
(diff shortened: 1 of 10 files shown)


Sources/Helium/HeliumUIKitViewController.swift
```diff
@@ -50,15 +73,23 @@ class HeliumViewController: UIViewController {
...
     override func viewDidDisappear(_ animated: Bool) {
         super.viewDidDisappear(animated)
-        if isBeingDismissed {
+        if isBeingDismissed || isMovingFromParent {
             HeliumPaywallPresenter.shared.cleanUpPaywall(heliumViewController: self)
+            presentationState.isOpen = false
         }
     }
     
+    @objc private func appWillTerminate() {
```
anishpdoshi, line 88: just want to make sure if someone has wired up `appWillTerminate` on their end, we'd still call it? i.e. that this notification -> appWillTerminate flow will still make sure any other termination flows they have will run

anishpdoshi: Gotcha cool!

PR #57, opened 2025-06-18
App account tokens!

salami:
Add new `heliumPurchase` method which should be used in place of the StoreKit 2 purchase method. It will automatically handle appAccountToken setting and create a new logging event so we can be sure to have a way to link the user/experiment up to purchase transactions.

Is the event overkill? Maybe... perhaps we just throw an error and don't allow custom appAccountToken, at least to start. And then strongly discourage custom Helium user ID's that are not a UUID.

For RevenueCat we will ask them to initialize with Helium user ID and strongly discourage setting a custom Helium user ID. RC will automatically set appAccountToken if the appUserId is a UUID, syncing things up for us.

### DIFF
--- a/Sources/Helium/HeliumCore/HeliumIdentityManager.swift
+++ b/Sources/Helium/HeliumCore/HeliumIdentityManager.swift
@@ -46,11 +46,15 @@ public class HeliumIdentityManager {
         self.heliumUserTraits = traits;
     }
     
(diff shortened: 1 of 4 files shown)


anishpdoshi (approved):
Lgtm! Not sure if there was more you wanted to add, approving to unblock

PR #58, opened 2025-06-23
Download retries

salami:


sjakati98 (approved):
Changes look good to me, exponential backoff should give us another try withoutbrowning out the server

PR #59, opened 2025-07-09
New helium delegates

salami:
StoreKitDelegate and RevenueCatDelegate

### DIFF
--- a/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
+++ b/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
@@ -100,13 +100,12 @@ public class StoreKitDelegate: HeliumPaywallDelegate {
     }
     
     public func restorePurchases() async -> Bool {
-        do {
-            try await AppStore.sync()
-            return true
-        } catch {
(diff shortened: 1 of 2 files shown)


Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
```diff
@@ -46,6 +46,88 @@ public extension HeliumPaywallDelegate {
...
+            }
+        } catch {
+            print("[Helium] StoreKitDelegate - Purchase failed with error: \(error.localizedDescription)")
+            return .failed(error)
+        }
+    }
+    
+    public func restorePurchases() async -> Bool {
+        do {
+            try await AppStore.sync()
```
anishpdoshi, line 104: @salami I think i actually might have messed up this in the docs previously - I'm not sure the AppStore.sync() and then immediate return true actually does anything. I ended up changing it to use Transaction.currentEntitlements - does this look more correct to you

   func restorePurchases() async -> Bool {
        for await result in Transaction.currentEntitlements {
            if case .verified = result {
                return true
            }
        }
        return false
    }

PR #62, opened 2025-07-24
Fix close event completely closed

salami:
The isOpen logic was just a tick ahead of full dismissal for the normal presentUpsell flow. This should take care of it by relying on the dismiss completion handler. It results in a bit of duplicate logic for open/close events but this is arguably easier to follow for the presentUpsell flow.

https://linear.app/tryhelium/issue/HEL-1297/fix-paywallclose-event-delay

### DIFF
--- a/Sources/Helium/HeliumPaywallPresenter.swift
+++ b/Sources/Helium/HeliumPaywallPresenter.swift
@@ -4,7 +4,14 @@ import UIKit
 
 class HeliumPaywallPresenter {
     static let shared = HeliumPaywallPresenter()
-    private init() {}
+    private init() {
+        NotificationCenter.default.addObserver(
+            self,
+            selector: #selector(appWillTerminate),
+            name: UIApplication.willTerminateNotification,
+            object: nil
+        )
+    }
     
     private var paywallsDisplayed: [HeliumViewController] = []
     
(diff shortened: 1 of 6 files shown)


Sources/Helium/HeliumPaywallPresenter.swift
```diff
@@ -4,26 +4,35 @@ import UIKit
 
 class HeliumPaywallPresenter {
     static let shared = HeliumPaywallPresenter()
-    private init() {}
+    private init() {
+        NotificationCenter.default.addObserver(
```
anishpdoshi, line 8: Does this mean we trigger a paywallClose event for all paywalls when the app terminates? Will it detect if an existing paywall is open?

Sources/Helium/HeliumPaywallPresenter.swift
```diff
@@ -58,16 +69,45 @@ class HeliumPaywallPresenter {
...
             return
         }
+        var paywallsRemoved = paywallsDisplayed
         Task { @MainActor in
             // Have the topmost paywall get dismissed by its presenter which should dismiss all the others,
             // since they must have ultimately be presented by the topmost paywall if you go all the way up.
-            paywallsDisplayed.first?.presentingViewController?.dismiss(animated: true, completion: onComplete)
+            paywallsDisplayed.first?.presentingViewController?.dismiss(animated: true) { [weak self] in
+                onComplete?()
+                self?.dispatchCloseForAll(paywallVCs: paywallsRemoved)
```
anishpdoshi, line 78: Just trying to understand the case where we have multiple paywall VCs and hence need to close on all. Is that for second try flows? Or are there other cases in which there'd be more than one paywall VC?

(Do we just create a different one per trigger?)

Sources/Helium/HeliumPaywallPresenter.swift
salami: Currently yes, it would try to trigger a paywallClose event for all paywalls in navigation stack, whether or not the paywall is visible view controller or not.

Sources/Helium/HeliumPaywallPresenter.swift
salami: It's for second try yes although theoretically could support more than that. So if user makes successful purchase or restore on a second try paywall, all paywalls should be closed here.

anishpdoshi (approved):
LGtm

PR #65, opened 2025-07-30
Testflight indication

salami:
https://linear.app/tryhelium/issue/HEL-954/log-if-app-is-in-testflight-vs-not

- Use the value from AppTransaction.shared if available (appStoreReceiptURL is deprecated)
- This should be fairly accurate, although it's hard to really test all the different scenarios

I think the main question is do we like these values?
`debug`
`sandbox`
`production`

Or should we remove `debug` and keep it simpler? Or even go more complex and add `simulator`, perhaps as a different field?

### DIFF
--- a/Sources/Helium/HeliumCore/AppReceiptsHelper.swift
+++ b/Sources/Helium/HeliumCore/AppReceiptsHelper.swift
@@ -12,8 +12,13 @@ class AppReceiptsHelper {
     static let shared = AppReceiptsHelper()
     
     private var appTransactionEnvironment: String? = nil
+    private var setupCompleted: Bool = false
     
     func setUp() {
+        if setupCompleted {
+            return
+        }
+        setupCompleted = true
         if #available(iOS 16.0, *) {
             Task {
                 let verificationResult = try? await AppTransaction.shared


Sources/Helium/HeliumPublic.swift
```diff
@@ -136,7 +138,8 @@ public class Helium {
         
         HeliumIdentityManager.shared.revenueCatAppUserId = revenueCatAppUserId
         
-        self.initialized = true;
+        AppReceiptsHelper.shared.setUp()
```
salami, line 141: Note that identify and helium_initializeCalled events are called shortly after this, which means they don't get the `AppTransaction.shared` result since that is async. In my testing this async result can take up to a second. We could potentially await before firing those 2 events? Otherwise it falls back to `appStoreReceiptURL`, which should still be fairly accurate.

anishpdoshi:
@salami  
- debug/sandbox/production sounds good to me - does sandbox = Testflight? 

I like separating debug from sandbox, simulator might be overkill

Sources/Helium/HeliumPublic.swift
anishpdoshi: I think if we `await` here it might block downloads - though maybe it could be parallelized?

If that's too hard, how it is right now (where identify/initializeCalled/potentially paywalldownloadsuccess) get hte non-deprecated route whereas the other events get the deprecated call method, seems fine to me. I don't think apple would get rid of the deprecated thing anytime soon...so should be safe.

anishpdoshi (approved):
Looks good to me!

PR #68, opened 2025-08-07
Fallback assets

salami:
A few key things here:
- This shows exactly what's in the html file so it's not localized.
- Localized prices will likely not work unless Helium.initialize was successful in which case a fallback probably not shown anyways.
- No special config stuff like background colors or actionConfig (whatever that is... is it even ever used?)

Note that the way this is coded now, open/dismiss/close events will fire... this has to be hashed out more. We might want to add a flag on these events or modify the paywall template name? I'll address this in a separate PR.

### DIFF
--- a/Sources/Helium/HeliumCore/HeliumBaseTemplateView.swift
+++ b/Sources/Helium/HeliumCore/HeliumBaseTemplateView.swift
@@ -32,6 +32,26 @@ public struct DynamicBaseTemplateView: BaseTemplateView {
         self.triggerName = trigger;
     }
     
+    public init(trigger: String, fallbackAsset: URL) {
+        // All that really matters here is paywallTemplateName
+        let fallbackPaywallInfo = HeliumPaywallInfo(paywallID: -1000, paywallTemplateName: "fallback_asset", productsOffered: [], resolvedConfig: "", shouldShow: true, fallbackPaywallName: "fallback_asset")
+        let delegate = HeliumActionsDelegate(paywallInfo: fallbackPaywallInfo, trigger: trigger)
+        _actionsDelegate = StateObject(wrappedValue: delegate)
+        _actionsDelegateWrapper = StateObject(wrappedValue: ActionsDelegateWrapper(delegate: delegate))
+        
+        // Provide fallback asset values
+        self.templateValues = JSON([
+            "baseStack" : [
+                "type" : "webView",
+                "name" : "webView",
+                "componentProps" : [
+                    "fallbackAssetURL" : fallbackAsset.absoluteString
+                ]
+            ]
+        ])
+        self.triggerName = trigger
+    }
+    
     public init(paywallInfo: HeliumPaywallInfo, trigger: String) {
         let delegate = HeliumActionsDelegate(paywallInfo: paywallInfo, trigger: trigger);
         _actionsDelegate = StateObject(wrappedValue: delegate)
--- a/Sources/Helium/HeliumCore/HeliumBaseTemplateView.swift
+++ b/Sources/Helium/HeliumCore/HeliumBaseTemplateView.swift
@@ -32,26 +32,6 @@ public struct DynamicBaseTemplateView: BaseTemplateView {
         self.triggerName = trigger;
     }
     
-    public init(trigger: String, fallbackAsset: URL) {
-        // All that really matters here is paywallTemplateName
-        let fallbackPaywallInfo = HeliumPaywallInfo(paywallID: -1000, paywallTemplateName: "fallback_asset", productsOffered: [], resolvedConfig: "", shouldShow: true, fallbackPaywallName: "fallback_asset")
-        let delegate = HeliumActionsDelegate(paywallInfo: fallbackPaywallInfo, trigger: trigger)
-        _actionsDelegate = StateObject(wrappedValue: delegate)
-        _actionsDelegateWrapper = StateObject(wrappedValue: ActionsDelegateWrapper(delegate: delegate))
-        
-        // Provide fallback asset values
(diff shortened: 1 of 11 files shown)


Sources/Helium/HeliumPublic.swift
```diff
@@ -98,30 +99,30 @@ public class Helium {
...
+    /// @param customUserTraits  Optional custom user traits for targeting
+    /// @param revenueCatAppUserId  Optional RevenueCat user ID for integration. Important if you are using RevenueCat to handle purchases!
+    /// @param fallbackAssetsConfig  (Optional) Provide html assets to use as fallback paywalls.
+    /// @param fallbackPaywallPerTrigger  Optional trigger-specific fallback views
+    ///
     public func initialize(
         apiKey: String,
         heliumPaywallDelegate: HeliumPaywallDelegate,
         fallbackPaywall: (any View),
-        baseTemplateView: (any BaseTemplateView.Type)? = nil,
```
salami, line 119: Thinking to remove this... I doubt anyone uses it?

anishpdoshi: Yeah i'm good with removing too... had this for Locket, but i think we should migrate them off of this
Don't think anyone else uses it, and if they do i'm fine with them just messaging in to migrate off

Sources/Helium/HeliumPublic.swift
```diff
@@ -98,30 +99,30 @@ public class Helium {
...
         triggers: [String]? = nil,
         customUserId: String? = nil,
         customAPIEndpoint: String? = nil,
         customUserTraits: HeliumUserTraits? = nil,
         revenueCatAppUserId: String? = nil,
+        fallbackAssetsConfig: FallbackAssetsConfig? = nil,
```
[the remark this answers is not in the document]
salami: fallbackBundleConfig ? or fallbackPaywallBundle/s? I think either works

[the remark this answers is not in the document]
anishpdoshi: fallbackBundleConfig sounds good! Since its a configuration object and we'll be loading JSON/typed values

Sources/Helium/HeliumCore/HeliumFallbackViewManager.swift
```diff
@@ -15,16 +15,41 @@ public class HeliumFallbackViewManager {
...
     
     // **MARK: - Public Methods**
-    public func setTriggerToFallback(toSet: [String: AnyView]) {
-        self.triggerToFallbackView = toSet
+    public func setFallbackBundleConfig(_ config: FallbackBundleConfig) {
+        fallbackBundleConfig = config
+        // Give immediate feedback if assets are not accessible & avoid trying to use later.
+        // This is synchronous but very fast (typically < 1 ms).
+        var fallbackAssetCount: Int = 0
+        var foundCount: Int = 0
```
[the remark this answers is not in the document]
salami: Makes sense! I'll adjust for this.

Sources/Helium/HeliumCore/HeliumBaseTemplateView.swift
salami, line 1: More unused code cleanup

Sources/Helium/HeliumPublic.swift
```diff
@@ -34,49 +39,63 @@ public class Helium {
...
+        } else {
+            return fallbackViewFor(trigger: trigger, templateName: paywallInfo?.paywallTemplateName)
+        }
+    }
+    
+    private func fallbackViewFor(trigger: String, templateName: String?) -> UpsellViewResult {
+        HeliumPaywallDelegateWrapper.shared.onHeliumPaywallEvent(event: .paywallOpenFailed(
+            triggerName: trigger,
+            paywallTemplateName: templateName ?? "Unknown"
+        ))
```
salami, line 81: The paywallOpenFailed event was only firing here if an error was thrown while attempting to display an actual paywall view, which is probably quite rare.

I moved it here so it will fire whenever a fallback is shown... but is that confusing? Should we just strip it out?

anishpdoshi: Interesting, stripping it out makes sense to me. 
I wonder if we should instead optimize for the following:
1) trying to make sure that in all classes of errors, we show a fallback paywall
2) when we do show a fallback paywall, add an error message/reason type field (maybe an enum?) that tells us what reason we're showing the fallback

Sources/Helium/HeliumCore/HeliumFallbackViewManager.swift
```diff
@@ -15,16 +16,47 @@ public class HeliumFallbackViewManager {
...
+                if let bundles = loadedConfig?.bundles, !bundles.isEmpty {
+                    try HeliumAssetManager.shared.writeBundles(bundles: bundles)
+                    print("[Helium] Successfully loaded paywalls from fallback bundle file.")
+                } else {
+                    print("[Helium] No bundles found in fallback bundle file.")
+                }
+            } catch {
+                print("[Helium] Failed to load fallback bundle: \(error)")
+            }
+        }
```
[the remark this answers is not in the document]
anishpdoshi: I wonder if we should actually be more comfortable making initialize "slow", because the download is already async, and someone can always wrap us in a Task {}/even a loading state themselves if they feel like its taking too long. But i guess it depends on the task difference, i.e. if its like a second diff between sync/async invocation of this block, then probably good to make it a task

Could also try a high priority task? https://developer.apple.com/documentation/swift/taskpriority/high

anishpdoshi (approved):
Thank you for all of the fixes/code cleanup as well!!!

PR #69, opened 2025-08-08
Custom app acct token

salami:
https://linear.app/tryhelium/issue/HEL-1488/updates-for-custom-app-store-server-app-attribution-token-changes

### DIFF
--- a/Sources/Helium/HeliumPublic.swift
+++ b/Sources/Helium/HeliumPublic.swift
@@ -109,6 +109,7 @@ public class Helium {
     * @param customUserId - Optional custom user ID to override default user identification
     * @param customAPIEndpoint - Optional custom API endpoint URL
     * @param customUserTraits - Optional custom user traits for targeting
+    * @param appAttributionToken - Optional Set this if you use a custom appAccountToken with your StoreKit purchases.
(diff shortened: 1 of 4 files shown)


Sources/Helium/HeliumPublic.swift
```diff
@@ -199,32 +213,32 @@ extension Product {
...
-        }
         
-        guard let appAccountToken = UUID(uuidString: HeliumIdentityManager.shared.getHeliumPersistentId()) else {
-            throw HeliumPurchaseError.invalidPersistentId
+        if let existingTokenOption {
+            let stringDescribingToken = String(describing: existingTokenOption)
+            if !stringDescribingToken.contains(appAccountToken.uuidString.lowercased()) && !stringDescribingToken.contains(appAccountToken.uuidString.uppercased()) {
+                throw HeliumPurchaseError.appAccountTokenMismatch
+            }
         }
```
salami, line 227: This logic is a bit sketchy although it seems to work in practice. For some reason String(describing:) has the uuid in lowercase but appAccountToken.uuidString is uppercase

salami, line 227: This check is kinda sketchy although it seems to work in practice. String(describing:) has uuid lowercase whereas appAccountToken.uuidString has it uppercase

PR #71, opened 2025-08-21
Fallback addendum (after server implemented)

salami:
- Remove `Kingfisher` dependency
  - RN doesn't like version Kingfisher 8.0 but figured we might as well just remove it... it's only used for `DynamicImage` which I don't believe is actually used at all?
- Remove fallback bundle data massaging since it'll be done on server
- If no analytics available (`Helium.initialize` was not successful) try using the data provided by fallback bundle so can still potentially log helium paywall events.
  - Is it safe to store analytics key/endpoint in the bundle? We could definitely strip if so. But I thought theoretically it might be useful to still try and log with fallback bundle values.

server piece here
https://github.com/cloudcaptainai/content-manager-lite/pull/96

### DIFF
--- a/Helium.podspec
+++ b/Helium.podspec
@@ -16,7 +16,6 @@ Pod::Spec.new do |s|
   
   s.source_files = ['Sources/Helium/**/*']
 
-  s.dependency 'Kingfisher', '~> 8.0'
   s.dependency 'AnyCodable-FlightSchool', '~> 0.6.0'
   s.dependency 'AnalyticsSwiftCocoapod', '~> 1.7'
   s.dependency 'SwiftyJSON', '~> 5.0.2'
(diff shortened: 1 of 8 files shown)


anishpdoshi (approved):

anishpdoshi:
yaay ty for removing Kingfisher that library is cool but I also remember causing a ton of random issues. DynamicImage and basically every component besides DynamicWebView and DynamicPositionedComponent can be removed, they were for the old native renderer which we don't prob need anymore

> If no analytics available (Helium.initialize was not successful) try using the data provided by fallback bundle so can still potentially log helium paywall events
Yes - this is great. analytics configs are set at a per app level in the frontend so this should be safe. though do wonder if there's a case where we're showing fallback in the middle of retrying?

PR #79, opened 2025-09-02
Remove native renderer for now

anishpdoshi:
The native renderer (non-webview code) is no longer active. We can always use Git if needed to change this, but this simplifies the SDK and should make it easier to port.


Sources/Helium/HeliumCore/HeliumBaseTemplateView.swift
```diff
@@ -26,25 +24,14 @@ public struct DynamicBaseTemplateView: BaseTemplateView {
...
-                    DynamicPositionedComponent(
-                        json: templateValues[modalScreenToShow],
-                        geometryProxy: reader
-                    )
-                }
-            }
-        }
+        // Directly use DynamicWebView
+        DynamicWebView(
+            json: templateValues,
```
salami, line 29: I think you may need to pass in templateVales["baseStack"]["componentProps"] here? I can test though and loop it into the release I'm planning to do in the next hour or two

PR #87, opened 2025-09-11
min ios 15 for SPM; remove mac version support

salami:


anishpdoshi (changes requested):
Looks like a few changes got added to this PR - The one I think is blocking is the paywall skipped firing event + if loadingStateEnabledFor is actually used

salami:
> Looks like a few changes got added to this PR - The one I think is blocking is the paywall skipped firing event + if loadingStateEnabledFor is actually used

Sorry I was building off of this branch for #88 and #89. loadingStateEnabledFor is used by RN/Expo. And I do want to change the paywall skipped event to a paywall open failed event, will create another PR for that shortly

anishpdoshi (approved):

PR #90, opened 2025-09-12
add isSecondTry to PaywallContextEvents

salami:
I think it will be useful in general but also:
- Only clear paywall event handlers when original paywall closes (note that second try open/close WILL call event handlers, but you can look at isSecondTry value)
- RN/Expo will be able to look at PaywallOpenFailedEvent and determine whether to call onFallback (if not a second try paywall)


anishpdoshi (approved):
Changes look great!
- I'd just make sure to update docs on the new events/fields, and what version of the SDK (maybe 3.0.1?) new event fields will be available in. But we could also group these kinds of doc updates per release, not sure
- Might be worth it to set up a separate "How to add a second try paywall" that end to end walks through the flow

PR #91, opened 2025-09-15
Rc id better

salami:
https://linear.app/tryhelium/issue/HEL-2028/support-post-setting-of-rc-id-after-initialization

- User can set rc app user id after init
- Make delegate update rc app user id
- Also set RC user attribute

### DIFF
--- a/Sources/HeliumRevenueCat/HeliumRevenueCat.swift
+++ b/Sources/HeliumRevenueCat/HeliumRevenueCat.swift
@@ -39,6 +39,9 @@ open class RevenueCatDelegate: HeliumPaywallDelegate {
             print("[Helium] RevenueCatDelegate - RevenueCat has not been configured. You must either configure it before initializing RevenueCatDelegate or pass in revenueCatApiKey to RevenueCatDelegate initializer.") 
         }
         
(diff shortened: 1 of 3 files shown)


Sources/HeliumRevenueCat/HeliumRevenueCat.swift
```diff
@@ -39,6 +41,15 @@ open class RevenueCatDelegate: HeliumPaywallDelegate {
...
         }
         
+        // Keep this value as up-to-date as possible
+        Helium.shared.setRevenueCatAppUserId(Purchases.shared.appUserID)
+        
+        if allowHeliumUserAttribute {
+            Purchases.shared.attribution.setAttributes([
+                "helium_hpid" : HeliumIdentityManager.shared.getHeliumPersistentId()
+            ])
+        }
```
salami, line 51: I realized these values are actually sent in RC webhooks. So we could potentially use this to link up events as well.

anishpdoshi: Amazing!

anishpdoshi (approved):

PR #92, opened 2025-09-16
New version 301

salami:
Prepping for new release

### DIFF
--- a/Sources/Helium/HeliumCore/BuildConstants.swift
+++ b/Sources/Helium/HeliumCore/BuildConstants.swift
@@ -8,5 +8,5 @@
  */
 public struct BuildConstants {
     /// Current SDK version
-    public static let version = "3.0.0"
+    public static let version = "3.0.1"
 }
--- a/Sources/Helium/HeliumPaywallPresenter.swift
+++ b/Sources/Helium/HeliumPaywallPresenter.swift
@@ -71,7 +71,7 @@ class HeliumPaywallPresenter {
             let triggerLoadingView = fallbackConfig.loadingView(for: trigger)
             
             // If loading state disabled for this trigger, show fallback immediately
-            if !useLoading {
+            if !useLoading || Helium.shared.getDownloadStatus() != .inProgress {
                 presentUpsell(trigger: trigger, from: viewController)
                 return
             }


Sources/Helium/HeliumCore/HeliumFetchedConfig.swift
```diff
@@ -98,6 +98,8 @@ public class HeliumFetchedConfigManager: ObservableObject {
         completion: @escaping (Result<HeliumFetchResult, Error>) -> Void
     ) {
         Task {
+            await self.updateDownloadState(.inProgress)
+
```
salami, line 102: It seems we weren't ever setting this

Sources/Helium/HeliumCore/StoreKitDelegate.swift
```diff
@@ -65,6 +65,10 @@ open class StoreKitDelegate: HeliumPaywallDelegate {
     open func onHeliumPaywallEvent(event: HeliumPaywallEvent) {
         // Override in a subclass if desired
     }
+    
+    open func onPaywallEvent(_ event: PaywallEvent) {
+        // Override in a subclass if desired
+    }
```
salami, line 71: Need to declare this so it can be subclassed

Sources/HeliumRevenueCat/HeliumRevenueCat.swift
```diff
@@ -6,7 +6,12 @@
 //
 
 import Helium
-import RevenueCat
+import class RevenueCat.Purchases
+import class RevenueCat.Offerings
+import class RevenueCat.CustomerInfo
+import class RevenueCat.StoreProduct
+import typealias RevenueCat.PurchaseResultData
+import class RevenueCat.Package
```
salami, line 14: I don't love this but the only alternative I'm thinking of is renaming our `PaywallEvent`. (RevenueCat has their own `PaywallEvent` 😤 that is publicly exposed through their sdk.)

It should be fine although if someone imports both Helium and RevenueCat in same file they may run into conflicts...

Sources/HeliumRevenueCat/HeliumRevenueCat.swift
```diff
@@ -102,7 +107,7 @@ open class RevenueCatDelegate: HeliumPaywallDelegate {
             }
             
             if result == nil {
-                let productToPurchase = try await Purchases.shared.products([productId])
+                let productToPurchase = await Purchases.shared.products([productId])
```
salami, line 110: This can't throw, so just addressing an xcode warning

Sources/HeliumRevenueCat/HeliumRevenueCat.swift
anishpdoshi: I'm not opposed to renaming it to minimize any chance of conflict. Maybe just HeliumPaywallEvent?

anishpdoshi (approved):
Lgtm!

PR #93, opened 2025-09-17
Transaction ids

salami:
Capturing those ids! (adding on to `subscriptionSucceeded` event)

The possible scenarios I see are:
- Customer uses one of our delegates
  - We get txn IDs; no delay added between makePurchase `.purchased` result and `subscriptionSucceeded` firing
- Customer uses custom delegate with SK1
  - Very minimal delay in my testing; worst case could delay a second or so
- Customer uses custom delegate with SK2
  - Can add a delay of a second or so... we can encourage users to implement `HeliumDelegateReturnsTransaction` to avoid this
- For RN/Flutter -- I haven't tested yet, but I'm hopeful that if they use RC, the transaction will be picked up quick since RC already has an extra delay after purchase, giving more time to process transaction. Alternatively we can grab the transaction ID right from RC and return without delay, just won't have original transaction ID.

Alternatively we could fire a separate `transactionProcessed` event to avoid any delays but I'm kinda thinking it's not worth the extra complexity.

### DIFF
--- a/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
+++ b/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
@@ -146,6 +146,7 @@ public class HeliumPaywallDelegateWrapper: ObservableObject {
         case .restored:
             self.fireEvent(PurchaseRestoredEvent(productId: productKey, triggerName: triggerName, paywallName: paywallTemplateName))
         case .purchased:
+            let transactionRetrievalStartTime: DispatchTime = DispatchTime.now()
             var transactionIds: TransactionIdPair? = nil
             if let transactionDelegate = delegate as? HeliumDelegateReturnsTransaction,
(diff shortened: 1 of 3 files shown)


salami:
> The code looks great. I do have a few behavioral worries:
> 
> * I do think the custom delegate with SK2 is actually gonna be fairly common (if we allow it). If there are delays associated with the subscriptionSucceeded, my worry is that it might block paywall closing, or post-purchase flows in general. Maybe its ok? But might be good for us to log the time taken if its a noticeable delay so we can decide if we want to make additional changes
> * If it is significant, might be worth looking into a separate transaction event or another way of exposing the last transaction (e.g. we expose a Helium.getLatestTransaction() thats async and just polls or something

Sounds good! Added time logging. Another option to consider which we've discussed before, is altering makePurchase to return transaction ids or transaction. This would have the added benefit of letting us get transaction ids of consumables that are marked finished, which we can't get from SK2 transaction methods unless they add a flag to their Info.plist.

PR #94, opened 2025-09-18
avoid fatalErrors; safer top most VC logic; bump version

salami:


Sources/Helium/HeliumCore/Segment/SegmentCore/Analytics.swift
```diff
@@ -85,7 +85,8 @@ class Analytics {
         if Self.isActiveWriteKey(configuration.values.writeKey) {
             // If you're hitting this in testing, it could be a memory leak, or something async is still running
             // and holding a reference.  You can use XCTest.waitUntilFinished(...) to wait for things to complete.
-            fatalError("Cannot initialize multiple instances of Analytics with the same write key")
+            print("Cannot initialize multiple instances of Analytics with the same write key")
```
anishpdoshi, line 88: Good call!

Sources/Helium/HeliumPublic.swift
```diff
@@ -128,8 +128,9 @@ public class Helium {
     }
     
     func upsellViewResultFor(trigger: String) -> UpsellViewResult {
-        if (!initialized) {
-            fatalError("Helium.shared.initialize() needs to be called before presenting a paywall. Please visit docs.tryhelium.com or message founders@tryhelium.com to get set up!");
+        if !initialized {
+            print("[Helium] Helium.shared.initialize() needs to be called before presenting a paywall. Please visit docs.tryhelium.com or message founders@tryhelium.com to get set up!")
```
anishpdoshi, line 132: Hm, what would happen in this case if someone hasn't called initialize? They wouldn't hav ea fallback so would the view result basically resolve to an empty view?

Wonder if we should make this a #warning so that we nudge people hard towards calling initialize()

anishpdoshi (approved):
LGTM modulo question about initialize()!

Sources/Helium/HeliumPublic.swift
salami: It will hit other logic that sees the nil view and avoids presenting and dispatches a OpenFailed event. I _think_ empty view paywalls should be nearly impossible with the latest sdk... unless user explicitly provides as a fallback.

PR #99, opened 2025-09-26
Loading update fix

salami:
Discovered this while testing Flutter
- We should fire the download complete event (which tells loading paywall to swap content) AFTER analytics set from download and paywall download success fires. In Flutter this situation crashes if fallback bundle set. In iOS it seems to be ok but analytics might use fallback / sometimes iOS seems to show fallback when it shouldn't.
- Also flutter-related, paywall open failed event should not fire until presented paywall is completely dismissed, otherwise flutter fallback view won't display

### DIFF
--- a/Sources/Helium/HeliumCore/HeliumController.swift
+++ b/Sources/Helium/HeliumCore/HeliumController.swift
@@ -108,7 +108,11 @@ public class HeliumController {
                         numAttempts: numFetchRequests
                     )
                 )
-                // Use the config as needed
+                
+                NotificationCenter.default.post(
+                    name: NSNotification.Name("HeliumConfigDownloadComplete"),
+                    object: nil
+                )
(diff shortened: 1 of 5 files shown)


Sources/Helium/HeliumCore/HeliumController.swift
```diff
@@ -108,7 +108,11 @@ public class HeliumController {
                         numAttempts: numFetchRequests
                     )
                 )
-                // Use the config as needed
+                
+                NotificationCenter.default.post(
+                    name: NSNotification.Name("HeliumConfigDownloadComplete"),
+                    object: nil
+                )
```
salami, line 115: It's just firing a bit later now, after analytics configured and after paywall download success event.

anishpdoshi (approved):
Lgtm

PR #100, opened 2025-09-26
App transaction Id

salami:
This is a potentially very powerful ID. Uniquer per user but same across their devices. Persists after app re-install.
https://developer.apple.com/documentation/storekit/apptransaction/apptransactionid

I discovered it about a week ago and think the sooner we can start capturing it the better.

### DIFF
--- a/Sources/Helium/HeliumCore/AppReceiptsHelper.swift
+++ b/Sources/Helium/HeliumCore/AppReceiptsHelper.swift
@@ -8,10 +8,16 @@
 import StoreKit
 
 class AppReceiptsHelper {
+
+    enum Environment: String {
+        case debug = "debug"
+        case sandbox = "sandbox"
+        case production = "production"
+    }
     
     static let shared = AppReceiptsHelper()
     
-    private var appTransactionEnvironment: String? = nil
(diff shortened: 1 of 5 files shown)


Sources/Helium/HeliumCore/AppReceiptsHelper.swift
```diff
@@ -20,10 +26,12 @@ class AppReceiptsHelper {
...
         if Bundle.main.appStoreReceiptURL != nil {
-            // Just rely on appStoreReceiptURL if available, even though it is deprecated.
             // AppTransaction.shared can trigger Apple account sign-in dialog in debug/sandbox
-            // if not signed into a sandbox account, which is annoying for sdk integrators.
-            return
+            // if not signed into a sandbox account, which is annoying for sdk integrators. So avoid
+            // that call if can determine sandbox from receipt.
+            if getEnvironment() == Environment.sandbox.rawValue {
+                return
+            }
```
salami, line 34: Initially I was a little skeptical about calling AppTransaction.shared in production for fear of sign-in dialogs, but I have an app that has been using AppTransaction.shared for months in production and haven't gotten any complaints.

anishpdoshi (approved):

PR #269, opened 2026-05-06
[HEL-5326] SDK pre-fetch path for Paddle web checkout

Arhan13:
## Summary

Adds the iOS-SDK side of the Paddle web-checkout pre-fetch optimization (HEL-5326). When an in-app paywall opens that has Paddle web products, the SDK fires bandit + Paddle BFF in the background per priceId so that by the time the user taps Subscribe, the responses are cached and the bundle in Safari can skip its own round-trips. User-perceived latency on click moves from "~2s of Safari loading" to "instant Safari open + Apple Pay sheet ready immediately."

## Architecture

```
Helium.upsell(trigger:)
  → HeliumPaywallPresenter.dispatchOpenEvent
       └─ if Paddle web products + paddleClientToken:
              PaddleCheckoutPrefetchCoordinator.shared.prefetch(priceIds: ..., …)
       └─ tasks run in background (parallel per priceId), user reads paywall

User taps Subscribe
  → ExternalWebCheckoutManager.startCheckoutFlow (Paddle path)
       └─ await PaddleCheckoutPrefetchCoordinator.shared.awaitOutcome(priceId:)
              ├─ .alreadyEntitled → invalidate cache, return .preCheckResolved (no Safari)
              ├─ .ready          → encodeBootstrapToCtx → ctx.paddleBootstrap
              └─ .failed/notStarted → no bootstrap (bundle does its own fetch — current behavior)

User dismisses paywall
  → HeliumPaywallPresenter.dispatchCloseEvent
       └─ PaddleCheckoutPrefetchCoordinator.shared.cancelAll()
```

The 1500ms hides behind the user's reading time. If the user clicks faster than the prefetch completes, `awaitOutcome` blocks until ready (per the design discussion: don't fall back to the slow path when prefetch is in-flight; just wait — Swift's `await task.value` returns instantly if already done).

## What's in this PR (6 commits, all RED→GREEN with TDD)

| # | Commit | What |
|---|---|---|
| 1 | `6feeec4` | Decode `paddleClientToken` from on-launch JSON into `HeliumFetchedConfig` |
| 2 | `610819c` | `HeliumPaymentAPIClient.createPaddleTransactionForPaywall` with structured 409 `duplicate_subscription` → `PaddlePrefetchError.alreadyEntitled` |
| 3 | `10bc486` | `PaddleBFFClient` — token-prefix URL selection (`test_*` → sandbox, otherwise prod), full raw-body preservation for forwarding to bundle |
| 4 | `139ff88` | `PaddleCheckoutPrefetchCoordinator` — `@MainActor`-isolated, per-priceId Task cache, `await`-or-instant on click, `cancelAll`/`cancelAllAndAwait` for paywall dismiss |
| 5+6 | `a4272d6` | Paywall lifecycle hooks (open → prefetch, close → cancel) + `ctx.paddleBootstrap` encoding in `buildEnrichedCheckoutURL` + click-time `awaitOutcome` branching in `startCheckoutFlow` |
| — | `299c7fa` | DRY cleanup pass — composite key parsing + URLRequest construction consolidated |

## Tests

**78/78 passing locally** via Mac Catalyst (Xcode 26.4, no iOS 26.4 simulator available so Mac Catalyst is the local test runner — exercises the same UIKit-bridged code paths). Helium-demo CI workflow will run the full iOS suite.

New test files:
- `OnLaunchResponseDecodingTests` — locks the on-launch JSON contract for `paddleClientToken`
- `PaddleCheckoutPrefetchClientTests` — 5 tests covering success path, request shape, 409 alreadyEntitled detection, non-duplicate 409 pass-through
- `PaddleBFFClientTests` — 9 tests covering URL selection per token prefix, headers, body shape, source_page derivation, response decoding, error path
- `PaddleCheckoutPrefetchCoordinatorTests` — 14 tests covering happy path, alreadyEntitled with BFF skip verification, both failure modes, cache misses, cancellation, parallel fan-out, ctx encoding, composite key extraction

New test infrastructure (reusable for future SDK API client tests):
- `MockURLProtocol` — URLSession interceptor with request capture + body re-materialization
- `HeliumPaymentAPIClient.init(urlSession:)` — internal DI initializer

## Behavior / safety

- **No-op when not configured** — if the on-launch response omits `paddleClientToken`, or the trigger has no `webProductsOfferedPaddle`, the prefetch never runs and the existing flow is unchanged.
- **Fail-open on transport errors** — `.failed` and `.notStarted` outcomes drop back to opening Safari without `ctx.paddleBootstrap`; the bundle does its own fetch (current behavior, no regression).
- **Cancellation on dismiss** — `dispatchCloseEvent` calls `cancelAll`; in-flight Tasks propagate cancellation into URLSession, no wasted bandit transactions for abandoned paywalls.
- **`alreadyEntitled` short-circuit** — Option A from the HEL-5326 design discussion: when bandit returns 409 `duplicate_subscription`, the SDK skips Safari entirely (rather than opening it just to see the bundle redirect to "already subscribed"), invalidates the entitlements cache, and returns `.preCheckResolved`.

## Dependencies

- **Bandit-side change** — needs cloudcaptainai/bandit-server-lite-phoenix#49 (`paddleClientToken` on `/on-launch`) merged to provide the field this SDK now consumes. Until that lands, `paddleClientToken` will be nil and the prefetch is a no-op (safe).
- **Bundler-side change** — separate PR will land on `arhan/hel-5326-paddle-sdk-prefetch-bundler` (in bundler-service) to actually consume `ctx.paddleBootstrap` and skip the bundle's own bandit + BFF calls. Until that lands, the SDK will produce the prefetched ctx but the bundle ignores it (also a no-op, no regression).

## Test plan

- [x] CI workflow passes against helium-demo
- [x] Manual end-to-end smoke once bandit PR #49 lands and we have `paddleClientToken` flowing — paywall display, observe prefetch logs, tap Subscribe, verify Apple Pay sheet appears faster than baseline
- [x] Cloudflare `__cf_bm` cookie behavior under SDK-originated step 1 + Safari-originated steps 4-6 — flagged as a known unknown in the design discussion; we'll find out empirically when we run against staging

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Touches web checkout URL construction and adds a new Paddle prefetch/caching pipeline, which could impact purchase flows if encoding, networking, or lifecycle hooks misbehave. Changes are guarded (no-op without `paddleClientToken`) and heavily tested, but failures would be user-facing at checkout time.
> 
> **Overview**
> Adds a Paddle web-checkout *prefetch* path: when a paywall opens, the SDK can now prefetch per-priceId Paddle transaction + checkout bootstrap data (and cancel it on paywall close), then inject that cached data into the browser checkout context to reduce click-to-checkout latency.
> 
> Updates `ExternalWebCheckoutManager.buildEnrichedCheckoutURL` to **compress** the JSON `ctx` (raw DEFLATE) and move it from the query string to the URL fragment, while ensuring `helium_ios_bundle_id` is present as a query param. Extends the payment API layer with URLSession DI, a Paddle `create-transaction-for-paywall` call with structured 409 "already entitled" handling, and adds a new Paddle BFF client plus comprehensive unit tests and test infrastructure (`MockURLProtocol`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 51d0a9a27bf9f16b6bd0bae6fa43ad981efe6dc0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Paddle prefetch during paywall presentation with short‑circuit for already‑entitled products.
  * Prefetched Paddle bootstraps can be embedded into the checkout URL context.
  * Launch config accepts an optional Paddle client token.
  * New Paddle BFF client and a session-aware prefetch coordinator with cancellation/timeouts.

* **Refactor**
  * Payment client networking now supports injected sessions and improved server‑error parsing.
  * Added optional ctx compression for checkout URLs.

* **Tests**
  * Extensive tests added for decoding, BFF, prefetch orchestration, compression, and mock networking.

### DIFF
--- a/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
+++ b/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
@@ -303,6 +303,19 @@ class HeliumPaywallDelegateWrapper {
             Task { @MainActor in
                 PaddleCheckoutManager.shared.stopObserving(paywallSession: paywallSession)
                 StripeCheckoutManager.shared.stopObserving(paywallSession: paywallSession)
+                PaddleCheckoutPrefetchCoordinator.shared.handlePaywallClose(paywallSession: paywallSession)
+            }
+        }
+
+        // Paddle SDK pre-fetch: kick off bandit + BFF for every Paddle web
+        // priceId on the trigger as soon as the paywall becomes visible, so
+        // the responses are cached by the time the user taps Subscribe.
+        // Hooked here (rather than in HeliumPaywallPresenter) so embedded
+        // paywall displays — which fire PaywallOpenEvent without going
+        // through presentUpsell — get the same optimization.
+        if event is PaywallOpenEvent, let paywallSession {
+            Task { @MainActor in
+                PaddleCheckoutPrefetchCoordinator.shared.handlePaywallOpen(paywallSession: paywallSession)
             }
         }
     }
--- a/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
+++ b/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
@@ -68,6 +68,14 @@ class HeliumPaywallDelegateWrapper {
         
         let transactionStatus: HeliumPaywallTransactionStatus
 
+        // Carry the prefetch's existingSubscriptionId (when present)
+        // through to the fired PurchaseRestoredEvent for analytics parity
+        // with the bundle's helium_purchase_already_entitled Jitsu fire's
+        // canonicalJoinTransactionId. Set only on the Paddle pre-checkout-
+        // entitled path (.preCheckResolved with a sub id); nil for
+        // StoreKit, Stripe, and cached-entitlements pre-checks.
+        var entitledExistingSubscriptionId: String? = nil
+
         let allStripeProductIds = Array((HeliumFetchedConfigManager.shared.getStripeProductsPriceMap() ?? [:]).keys)
         let stripeApplePayFlowEnabled = ApplePayHelper.shared.getStripeApplePayAvailable()
 
@@ -84,35 +92,46 @@ class HeliumPaywallDelegateWrapper {
 
         if paymentProcessor == .stripe && !stripeApplePayFlowEnabled {
             do {
-                transactionStatus = try await StripeCheckoutManager.shared.startCheckoutFlow(
+                let outcome = try await StripeCheckoutManager.shared.startCheckoutFlow(
                     for: productKey,
                     triggerName: triggerName,
                     paywallSession: paywallSession
-                ).transactionStatus
+                )
+                transactionStatus = outcome.transactionStatus
+                entitledExistingSubscriptionId = outcome.existingSubscriptionId
             } catch {
                 transactionStatus = .failed(error)
             }
         } else if paymentProcessor == .paddle {
             do {
-                transactionStatus = try await PaddleCheckoutManager.shared.startCheckoutFlow(
+                let outcome = try await PaddleCheckoutManager.shared.startCheckoutFlow(
                     for: productKey,
                     triggerName: triggerName,
                     paywallSession: paywallSession
-                ).transactionStatus
+                )
+                transactionStatus = outcome.transactionStatus
+                entitledExistingSubscriptionId = outcome.existingSubscriptionId
             } catch {
                 transactionStatus = .failed(error)
             }
         } else {
             transactionStatus = await delegate.makePurchase(productId: productKey)
         }
-        
+
         switch transactionStatus {
         case .cancelled:
             self.fireEvent(PurchaseCancelledEvent(productId: productKey, triggerName: triggerName, paywallName: paywallTemplateName, paymentProcessor: paymentProcessor), paywallSession: paywallSession)
         case .failed(let error):
             self.fireEvent(PurchaseFailedEvent(productId: productKey, triggerName: triggerName, paywallName: paywallTemplateName, error: error, paymentProcessor: paymentProcessor), paywallSession: paywallSession)
         case .restored:
-            self.fireEvent(PurchaseRestoredEvent(productId: productKey, triggerName: triggerName, paywallName: paywallTemplateName, restoreOrigin: .duringPurchase, paymentProcessor: paymentProcessor), paywallSession: paywallSession)
+            self.fireEvent(PurchaseRestoredEvent(
+                productId: productKey,
+                triggerName: triggerName,
+                paywallName: paywallTemplateName,
+                restoreOrigin: .duringPurchase,
+                paymentProcessor: paymentProcessor,
+                existingSubscriptionId: entitledExistingSubscriptionId
+            ), paywallSession: paywallSession)
         case .purchased:
             let transactionRetrievalStartTime: DispatchTime = DispatchTime.now()
             var transactionIds: HeliumTransactionIdResult? = nil
--- a/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
+++ b/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
@@ -65,16 +65,8 @@ class HeliumPaywallDelegateWrapper {
         } ?? false
         
         StoreKit1Listener.ensureListening()
-        
-        let transactionStatus: HeliumPaywallTransactionStatus
 
-        // Carry the prefetch's existingSubscriptionId (when present)
-        // through to the fired PurchaseRestoredEvent for analytics parity
-        // with the bundle's helium_purchase_already_entitled Jitsu fire's
-        // canonicalJoinTransactionId. Set only on the Paddle pre-checkout-
-        // entitled path (.preCheckResolved with a sub id); nil for
-        // StoreKit, Stripe, and cached-entitlements pre-checks.
-        var entitledExistingSubscriptionId: String? = nil
+        let transactionStatus: HeliumPaywallTransactionStatus
 
         let allStripeProductIds = Array((HeliumFetchedConfigManager.shared.getStripeProductsPriceMap() ?? [:]).keys)
         let stripeApplePayFlowEnabled = ApplePayHelper.shared.getStripeApplePayAvailable()
@@ -98,7 +90,6 @@ class HeliumPaywallDelegateWrapper {
                     paywallSession: paywallSession
                 )
                 transactionStatus = outcome.transactionStatus
-                entitledExistingSubscriptionId = outcome.existingSubscriptionId
             } catch {
                 transactionStatus = .failed(error)
             }
@@ -110,7 +101,6 @@ class HeliumPaywallDelegateWrapper {
                     paywallSession: paywallSession
                 )
                 transactionStatus = outcome.transactionStatus
-                entitledExistingSubscriptionId = outcome.existingSubscriptionId
             } catch {
                 transactionStatus = .failed(error)
             }
@@ -129,8 +119,7 @@ class HeliumPaywallDelegateWrapper {
                 triggerName: triggerName,
                 paywallName: paywallTemplateName,
                 restoreOrigin: .duringPurchase,
-                paymentProcessor: paymentProcessor,
-                existingSubscriptionId: entitledExistingSubscriptionId
+                paymentProcessor: paymentProcessor
             ), paywallSession: paywallSession)
         case .purchased:
             let transactionRetrievalStartTime: DispatchTime = DispatchTime.now()
--- a/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
+++ b/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
@@ -315,12 +315,6 @@ class HeliumPaywallDelegateWrapper {
             }
         }
 
-        // Paddle SDK pre-fetch: kick off bandit + BFF for every Paddle web
-        // priceId on the trigger as soon as the paywall becomes visible, so
-        // the responses are cached by the time the user taps Subscribe.
-        // Hooked here (rather than in HeliumPaywallPresenter) so embedded
-        // paywall displays — which fire PaywallOpenEvent without going
-        // through presentUpsell — get the same optimization.
         if event is PaywallOpenEvent, let paywallSession {
             Task { @MainActor in
                 PaddleCheckoutPrefetchCoordinator.shared.handlePaywallOpen(paywallSession: paywallSession)
--- a/Sources/Helium/HeliumCore/StripeCheckout/CtxCompression.swift
+++ b/Sources/Helium/HeliumCore/StripeCheckout/CtxCompression.swift
@@ -25,8 +25,11 @@ enum CtxCompression {
     ///
     /// Returns nil when the framework reports a 0-byte output, which it
     /// does on encoder failure or when the destination buffer is too
-    /// small. Callers default to sending uncompressed `?ctx=` in that
-    /// case (no regression).
+    /// small. The bundler exclusively decodes compressed ctx, so
+    /// callers must abort the checkout on nil rather than emit an
+    /// uncompressed payload — the bundler would silently parse an
+    /// uncompressed payload as empty defaults and the user would
+    /// land on a paywall with no product data.
     static func deflateRaw(_ data: Data) -> Data? {
         // The encoder doesn't know the output size up-front, so we have
         // to allocate a destination buffer big enough for the worst case.
--- a/Sources/Helium/HeliumCore/StripeCheckout/CtxCompression.swift
+++ b/Sources/Helium/HeliumCore/StripeCheckout/CtxCompression.swift
@@ -1,42 +1,13 @@
 import Foundation
 import Compression
 
-/// Compresses ctx bytes (post-JSON-serialization, pre-base64URL) so the
-/// bundler URL doesn't blow Safari's URL budget when paywalls carry
-/// multiple pre-fetched Paddle bootstraps.
-///
-/// **Wire pairing:** Swift `COMPRESSION_ZLIB` ↔ JS
-/// `DecompressionStream('deflate-raw')`. Apple's `COMPRESSION_ZLIB`
-/// emits raw DEFLATE (RFC 1951, no zlib/gzip wrapper) despite the name —
-/// see Apple's docs for `compression.h`. `DecompressionStream('deflate-raw')`
-/// in browsers (Safari 16.4+) consumes raw DEFLATE exactly. Both ends
-/// are native, no third-party libraries on either side.
-///
-/// `CtxCompression.deflateRaw(...)` is intentionally a free function
-/// rather than a class method — it has no state, no actor isolation
-/// constraint, and the call site (`buildEnrichedCheckoutURL`) just
-/// needs `Data → Data?`.
 enum CtxCompression {
 
-    /// Apple's `compression_encode_buffer` doesn't expose a quality/level
-    /// parameter — the level is internal to the framework. For our use
-    /// (small JSON ctx, run once per Subscribe-tap), the default level
-    /// is fine; the size win comes from the algorithm, not the level.
-    ///
-    /// Returns nil when the framework reports a 0-byte output, which it
-    /// does on encoder failure or when the destination buffer is too
-    /// small. The bundler exclusively decodes compressed ctx, so
-    /// callers must abort the checkout on nil rather than emit an
-    /// uncompressed payload — the bundler would silently parse an
-    /// uncompressed payload as empty defaults and the user would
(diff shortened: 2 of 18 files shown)


Arhan13:
@coderabbitai review

Arhan13:
@coderabbitai review

Sources/Helium/HeliumPaywallPresentation/HeliumPaywallPresenter.swift
salami, line 1: Can we move the open/close call logic to fireEvent of HeliumPaywallDelegate? That way embedded paywall displays will work too. (HeliumPaywallPresenter only applies to the more common presentPaywall flow.) There is already existing logic there that acts on paywall-related events.

And also maybe move more of the logic to PaddleCheckoutPrefetchCoordinator so HeliumPaywallDelegate can call one-liners on the open/close events?

Sources/Helium/HeliumCore/StripeCheckout/ExternalWebCheckoutManager.swift
```diff
@@ -92,19 +92,52 @@ public class ExternalWebCheckoutManager: NSObject {
...
+        var paddleBootstrapDict: [String: Any]? = nil
+        if provider.providerSlug == "paddle",
+           let priceId = PaddleCheckoutPrefetchCoordinator.extractPriceId(from: productKey) {
+            let outcome = await PaddleCheckoutPrefetchCoordinator.shared.awaitOutcome(priceId: priceId)
+            switch outcome {
+            case .alreadyEntitled(let code, let message):
+                HeliumLogger.log(.debug, category: .entitlements,
+                                 "\(provider.displayName) prefetch alreadyEntitled (\(code)): \(message) — skipping browser")
+                // Mirror the pre-checkout entitlement-already-owns short-circuit
+                // (line ~185 above): no browser open, refresh entitlements so
```
salami: Please do a review of the opus 4.7-style comments. specific line references (this one does not seem accurate) and whatnot just add maintenance burden

Sources/Helium/HeliumPaywallPresentation/HeliumPaywallPresenter.swift
```diff
@@ -456,11 +456,57 @@ class HeliumPaywallPresenter {
...
         dispatchOpenOrCloseEvent(openEvent: false, paywallVC: paywallVC)
-        
+
+        // Cancel any in-flight prefetches scoped to THIS paywall's priceIds
+        // (HEL-5326). Bugbot review: previously called the global
+        // `cancelAll()`, which would wipe other concurrently-displayed
+        // paywalls' prefetch results too. Scoping by priceId leaves other
+        // paywalls' caches intact while still freeing this paywall's
+        // in-flight bandit + BFF tasks (avoiding wasted Paddle draft
+        // transactions when the user dismisses without clicking).
```
salami: Referring to specific bugbot review doesn't make sense to readers of codebase going forward

Sources/Helium/HeliumCore/StripeCheckout/PaddleCheckoutPrefetchCoordinator.swift
```diff
@@ -0,0 +1,401 @@
+import Foundation
+
+/// Outcomes the prefetch coordinator can produce for a given priceId.
+///
+/// Stage 5 (the click handler in ExternalWebCheckoutManager) pattern-matches
```
salami: stage 5 what does that mean?

Sources/Helium/HeliumCore/StripeCheckout/HeliumPaymentAPIClient.swift
```diff
@@ -78,6 +108,79 @@ public class HeliumPaymentAPIClient {
...
+    ///     style "product_id:price_id" composite that `baseRequestBody`'s
+    ///     `productId:` parameter produces. We bypass that parameter and set
+    ///     `priceId` explicitly — sending the wrong field name (or under
+    ///     `productPriceId`) makes bandit reject with 400 "priceId is required".
+    ///   - `orgId` (optional, fallback to API key): added for parity with
+    ///     the bundler's runtime call; bandit can resolve it from the API
+    ///     key but sending it explicitly is one less DB round-trip there.
+    ///   - identity fields (apiKey, userId, rcUserId, heliumPersistentId,
+    ///     appTransactionId, paddleCustomerId): unchanged, from baseRequestBody.
+    func createPaddleTransactionForPaywall(
```
salami, line 103: nit - this very long comment could be simplified. the code reads well enough on its own

salami (commented):
Overall looks great! Just added some nitpicky notes about the code comments, and one important note about the open/close logic that should be addressed

Sources/Helium/HeliumCore/StripeCheckout/ExternalWebCheckoutManager.swift
```diff
@@ -92,19 +92,56 @@ public class ExternalWebCheckoutManager: NSObject {
...
+                entitlementsSource.invalidateCache()
+                return .preCheckResolved
+            case .ready:
+                paddleBootstrapDict = PaddleCheckoutPrefetchCoordinator.encodeBootstrapToCtx(
+                    outcome,
+                    priceId: priceId
+                )
+            case .failed, .notStarted:
+                paddleBootstrapDict = nil // default to bundle's own fetch
+            }
```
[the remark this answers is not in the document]
salami: I think this is desired. It keeps PurchaseAlreadyEntitledEvent reserved for ios "re-purchase" events

Sources/Helium/HeliumCore/StripeCheckout/CtxCompression.swift
```diff
@@ -0,0 +1,58 @@
...
+
+    /// Apple's `compression_encode_buffer` doesn't expose a quality/level
+    /// parameter — the level is internal to the framework. For our use
+    /// (small JSON ctx, run once per Subscribe-tap), the default level
+    /// is fine; the size win comes from the algorithm, not the level.
+    ///
+    /// Returns nil when the framework reports a 0-byte output, which it
+    /// does on encoder failure or when the destination buffer is too
+    /// small. Callers default to sending uncompressed `?ctx=` in that
+    /// case (no regression).
```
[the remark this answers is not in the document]
Arhan13: The doc was the bug, not the code. Fixed in 032360e: updated `CtxCompression.deflateRaw`'s doc comment to match the actual throwing contract.

Why throwing is correct here (and uncompressed-fallback would be wrong): the bundler-service side now exclusively decodes compressed `?ctx=` (`decodeCompressedCtxParam` in `heliumStandalonePaddle.ts`). The legacy uncompressed path was intentionally removed in the coordinated bundler PR — these two PRs ship together and there is no merged SDK version that emits uncompressed ctx. So an uncompressed payload would be silently parsed as empty defaults by the bundler, dropping the user onto a paywall with no product data. The throw propagates up through `startCheckoutFlow`'s `try` chain to the SDK error boundary which emits a failure event — no host-app crash.

Resolving as the doc is now consistent with the throwing behavior.

Sources/Helium/HeliumCore/StripeCheckout/ExternalWebCheckoutManager.swift
Arhan13: Cleaned up — verified no specific line references (`(line ~185 above)`, `L###`, `Line N`, etc.) remain in `ExternalWebCheckoutManager.swift`, `HeliumPaymentAPIClient.swift`, `PaddleCheckoutPrefetchCoordinator.swift`, or `HeliumPaywallPresenter.swift`. The original culprit at line 112 (`// (line ~185 above): no browser open…`) is gone. Resolving.

Broader "trim verbose comments" follow-ups continue under the specific per-comment threads (e.g. the `createPaddleTransactionForPaywall` doc nit).

Sources/Helium/HeliumPaywallPresentation/HeliumPaywallPresenter.swift
Arhan13: Bugbot reference removed — `grep -n bugbot Sources/Helium/HeliumPaywallPresentation/HeliumPaywallPresenter.swift` returns nothing. Resolving.

Sources/Helium/HeliumCore/StripeCheckout/PaddleCheckoutPrefetchCoordinator.swift
Arhan13: "stage" wording removed — `grep -ni stage Sources/Helium/HeliumCore/StripeCheckout/PaddleCheckoutPrefetchCoordinator.swift` returns nothing. Resolving.

Sources/Helium/HeliumPaywallPresentation/HeliumPaywallPresenter.swift
Arhan13: Done — moved the open/close hooks into `HeliumPaywallDelegate.fireEvent` (commit lives in earlier on-branch work; visible at `Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift:321-339`):

```swift
if event is PaywallCloseEvent, let paywallSession {
    Task { @MainActor in
        ...
        PaddleCheckoutPrefetchCoordinator.shared.handlePaywallClose(paywallSession: paywallSession)
    }
}

// Hooked here (rather than in HeliumPaywallPresenter) so embedded
// paywall displays — which fire PaywallOpenEvent without going
// through presentUpsell — get the same optimization.
if event is PaywallOpenEvent, let paywallSession {
    Task { @MainActor in
        PaddleCheckoutPrefetchCoordinator.shared.handlePaywallOpen(paywallSession: paywallSession)
    }
}
```

And per your second suggestion, the coordinator now exposes one-liner entry points so the delegate doesn't pull config/extraction logic into itself: `handlePaywallOpen(paywallSession:)` reads `paywallSession.paywallInfoWithBackups.webProductsOfferedPaddle` and `paddleClientToken` from `HeliumFetchedConfigManager`, extracts priceIds, and calls `prefetch(...)`. `handlePaywallClose(paywallSession:)` is literally `cancelForSession(sessionId: paywallSession.sessionId)`. `HeliumPaywallPresenter` no longer references the coordinator at all (`grep -n PaddleCheckoutPrefetchCoordinator Sources/Helium/HeliumPaywallPresentation/HeliumPaywallPresenter.swift` returns nothing).

Resolving.

Sources/Helium/HeliumCore/StripeCheckout/HeliumPaymentAPIClient.swift
Arhan13: Trimmed in 3606c89 — went from 38 lines / 283 words → 13 lines / 93 words. Kept only what's not visible from the signature: the typed `alreadyEntitled` error path (so callers know to catch it) and the `pri_xxx`-vs-composite footgun (so future maintainers don't get bitten by bandit's "priceId is required" 400). Dropped: error-case enumeration (visible from the throw sites), the "why not `post()`" justification (visible from the manual decode), and the body-shape walkthrough (the request building is right there). Also fixed a stray "fallback to API key" parenthetical that was a CLAUDE.md violation. Resolving.

Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
```diff
@@ -84,35 +92,46 @@ class HeliumPaywallDelegateWrapper {
...
             self.fireEvent(PurchaseFailedEvent(productId: productKey, triggerName: triggerName, paywallName: paywallTemplateName, error: error, paymentProcessor: paymentProcessor), paywallSession: paywallSession)
         case .restored:
-            self.fireEvent(PurchaseRestoredEvent(productId: productKey, triggerName: triggerName, paywallName: paywallTemplateName, restoreOrigin: .duringPurchase, paymentProcessor: paymentProcessor), paywallSession: paywallSession)
+            self.fireEvent(PurchaseRestoredEvent(
+                productId: productKey,
+                triggerName: triggerName,
+                paywallName: paywallTemplateName,
+                restoreOrigin: .duringPurchase,
+                paymentProcessor: paymentProcessor,
+                existingSubscriptionId: entitledExistingSubscriptionId
```
salami: I'd suggest we do not add this existingSubscriptionId parameter. It's not used by our metrics and adds a Paddle-specific column, and I don't really see the value it provides. Or if we do add it, re-use the canonicalJoinTransactionId field so we don't add a new column.

There's a separate argument on whether all "already entitled" events should just be "restored" events, but don't think that's worth addressing at this time.

Arhan13: Done — went with Option A (drop the parameter entirely). Pushed in `fbdf935`.

Removed:
- `PurchaseRestoredEvent.existingSubscriptionId` field, init param, and the conditional `toDictionary` emission
- `HeliumPaywallDelegate.handlePurchase`'s `entitledExistingSubscriptionId` plumbing variable + the two `outcome.existingSubscriptionId` reads + the kwarg on the `PurchaseRestoredEvent` init
- `WebCheckoutOutcome.preCheckResolved`'s associated value (`existingSubscriptionId: String?`) and the `var existingSubscriptionId` accessor — they only existed to feed the field we just dropped
- `tappedShortCircuit` 3-tuple → 2-tuple (the sub-id slot only existed to plumb to the now-removed event field)
- 3 `EventToDictionaryTests` and 2 `tappedShortCircuit` tests that asserted the dropped field's behavior

Kept (still needed for the bundle's own `helium_purchase_already_entitled` Jitsu fire to include the sub id under `canonicalJoinTransactionId` as it always has):
- `PaddlePrefetchError.alreadyEntitled`'s `existingSubscriptionId`
- `WebCheckoutOutcome.alreadyEntitled`'s `existingSubscriptionId`
- `HeliumPaymentAPIClient.extractExistingSubscriptionId` 5-path lookup + tests
- `encodeAlreadyEntitledToCtx` sub-id passthrough into `ctx.paddleAlreadyEntitled`

Re: the broader "should `alreadyEntitled` events just be `restored` events" thread — agree it's worth a separate conversation. Resolving this one.

salami (approved):

Sources/Helium/HeliumCore/StripeCheckout/ExternalWebCheckoutManager.swift
```diff
@@ -199,8 +199,28 @@ public class ExternalWebCheckoutManager: NSObject {
...
+    /// **fragment** (after `#`), not the query string.
+    ///
+    /// **Why fragment, not query (HEL-5326 PII follow-up):** the ctx
+    /// payload contains PII — `customer.email` (Paddle's record of a
+    /// returning customer's email) and `ip_geo_postal_code` (Paddle's
+    /// IP-derived postal code) per the Paddle privacy review.
+    /// Hash fragments are **not sent in HTTP requests by browsers**, so
+    /// they don't appear in CDN access logs, server access logs, or
+    /// referer headers — exactly the leakage Paddle's privacy policy
+    /// avoids.
```
salami: We should avoid this explanation, this repo is public

Sources/Helium/HeliumCore/StripeCheckout/ExternalWebCheckoutManager.swift
```diff
@@ -277,16 +297,50 @@ public class ExternalWebCheckoutManager: NSObject {
...
+        // — Paddle's returning-customer email and IP-derived postal
+        // code) doesn't leak through CDN logs, server access logs, or
+        // referer headers. The bundler reads `ctx` exclusively from
+        // `window.location.hash`; no query-string wire is supported.
+        //
+        // base64URL alphabet (RFC 4648 §5: A-Z, a-z, 0-9, -, _) is
+        // entirely within RFC 3986's `unreserved` character set, so
+        // setting `components.fragment` directly is safe — no
+        // percent-encoding is applied to characters that don't need
+        // it.
```
salami: same here. The amount of comment to code here is a bit much, claude 4.7 going wild with comments lol

Tests/helium-swiftTests/CtxCompressionTests.swift
```diff
@@ -159,7 +159,188 @@ final class CtxCompressionTests: XCTestCase {
...
+    /// the compressed ctx to the URL **fragment** (after `#`), NOT the
+    /// query string. The ctx contains PII (`customer.email`,
+    /// `ip_geo_postal_code`) and hash fragments aren't sent in HTTP
+    /// requests by browsers — so the payload doesn't appear in CDN
+    /// logs, server access logs, or referer headers.
+    ///
+    /// This test locks in the wire move: ctx MUST be in the fragment,
+    /// MUST NOT be in the query, and any pre-existing query items on
+    /// the bundle URL are preserved (we don't accidentally clobber
+    /// `helium_ios_bundle_id` or any other bundle URL params).
```
salami: and here

PR #287, opened 2026-06-17
HEL-5430 Gate California (AB 2863) checkout block behind a server-config kill switch

Arhan13:
## Summary

The SDK currently **hard-blocks California buyers** before the Paddle web paywall opens. The bundler-service web paywall now shows an **AB 2863 affirmative-consent modal** to California buyers (detected via Paddle `ip_geo_postal_code`, US ZIP 90001–96162) and lets them complete the purchase *with* consent. This SDK is the upstream layer that does the hard block, so it needs to stop — but only behind a **server-side kill switch**, because the SDK is binary-shipped.

This PR makes the California block flippable via a remote flag that **defaults to the current blocking behavior**.

## What changed

- **New remote flag** `HeliumPaywallInfo.allowCaliforniaWebCheckout` — a computed property reading `additionalPaywallFields["allowCaliforniaWebCheckout"]`, sitting right next to `webPaywallBundleUrl`. Uses SwiftyJSON's strict `.bool` (not coercive `.boolValue`), so **only a literal JSON `true` flips the switch**; absent / null / wrong-typed / string `"true"` / number all default to `false` (block).
- **Threaded** `allowCaliforniaCheckout` through `PaddleCheckoutPrefetchCoordinator.prefetch` → `runPrefetchChain` (default `false`).
- **Gated the block** at `runPrefetchChain`: `if !allowCaliforniaCheckout, let caPostal = californiaPostalCode(...)`. Flag on ⇒ the prefetch resolves `.ready` and the buyer reaches the (consent-aware) paywall.
- `handlePaywallOpen` forwards `info.allowCaliforniaWebCheckout` into the prefetch.

### Intentionally left inert (not deleted)

`PaddleCaliforniaBlocked`, `californiaPostalCode`, `californiaBlockedPostalCode`, `WebCheckoutError.californiaBuyerBlocked`, and the `.caBlocked` telemetry arm are kept. They remain the **active mechanism for the flag-off (default) path**. When the flag is on, the coordinator never produces a `PaddleCaliforniaBlocked` outcome, so the consumer's throw in `ExternalWebCheckoutManager` is simply never reached. Removing them would risk the default-block behavior; they can be retired later when fully flag-on. The `ip_geo` observability and the `paddleResponseFixtureWithCruft` fixture (which proves `ip_geo_postal_code` survives into the bundle ctx) are kept.

## Tests (TDD)

Added behavioral tests (there were previously **no** behavioral tests asserting the block):

**`PaddleCheckoutPrefetchCoordinatorTests`**
- CA-ZIP prefetch resolves `.ready` when the flag is **on**
- CA-ZIP prefetch stays `.failed(PaddleCaliforniaBlocked)` when the flag is **off** (default)
- Non-CA US ZIP (NY 10001) resolves `.ready` even with the flag off (block stays CA-specific)
- End-to-end: `handlePaywallOpen` forwards the flag from paywall-info config → `.ready` when on, blocked when absent

**`OnLaunchResponseDecodingTests`** (JSON contract)
- `allowCaliforniaWebCheckout` decodes `true` from `additionalPaywallFields`
- defaults `false` when absent
- is `false` when explicitly `false`

Full suite: **135 tests, 0 failures** (run via Mac Catalyst locally; CI runs the iOS-simulator suite in helium-demo).

## ⚠️ Compliance & rollout

The SDK is **binary-shipped** — old versions keep blocking for weeks–months as adoption rolls out, and there's no way to retroactively change them. This is exactly why the kill switch defaults to **block**.

**Do not flip `allowCaliforniaWebCheckout` to `true` server-side until the bundler consent modal + bandit changes are confirmed live end-to-end.** Until then, every layer keeps the current safe (blocking) behavior with zero change in production.

## Cross-repo contract

Server sets, per-paywall, in the paywall info's `additionalPaywallFields`:
```json
{ "allowCaliforniaWebCheckout": true }
```
Must be a JSON boolean. Same `ip_geo_postal_code` field is still forwarded into the bundle ctx, so the bundle keeps detecting CA after the SDK stops blocking.

Linear: HEL-5430

---

> [!NOTE]
> **Medium Risk**
> Changes who can proceed to web checkout for California IP-geo buyers when the server flag is enabled; default-off and strict boolean parsing limit accidental unblock, but incorrect server rollout could affect AB 2863 compliance.
> 
> **Overview**
> Adds a **remote kill switch** so California (AB 2863) buyers are no longer hard-blocked during Paddle prefetch only when the server explicitly allows it; **default stays block** for older SDK builds and missing config.
> 
> **`HeliumPaywallInfo.allowCaliforniaWebCheckout`** reads `additionalPaywallFields["allowCaliforniaWebCheckout"]` with strict JSON boolean semantics (only literal `true` enables; absent/null/wrong types → `false`). **`PaddleCheckoutPrefetchCoordinator`** threads `allowCaliforniaCheckout` from `handlePaywallOpen` through prefetch into **`runPrefetchChain`**, where the existing CA ZIP check runs only when the flag is off; when on, prefetch returns **`.ready`** so buyers reach the consent-aware web paywall.
> 
> Tests lock the JSON contract (fail-closed decoding) and prefetch behavior (flag on/off, ZIP boundaries, non-CA US, end-to-end `handlePaywallOpen` wiring).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fca18c7eb088f9af10f6ffa67f6f0cc4ee8a0e94. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added configuration control for California web checkout availability

* **Tests**
  * Added comprehensive test coverage for California checkout scenarios, including edge cases and conditional behavior validation

### DIFF
--- a/Sources/Helium/HeliumCore/Models.swift
+++ b/Sources/Helium/HeliumCore/Models.swift
@@ -92,7 +92,17 @@ public struct HeliumPaywallInfo: Codable {
         }
         return nil
     }
-    
+
+    /// Server-config kill switch for the California (AB 2863) checkout block.
+    /// When the web bundle's affirmative-consent flow is confirmed live, the
+    /// server sets this true and California buyers are allowed through to the
+    /// (consent-aware) web paywall instead of being hard-blocked upstream.
+    /// Defaults to `false` (block) so binary-shipped older SDK versions — and any
+    /// config that omits the field — keep the safety net until the server flips it.
+    var allowCaliforniaWebCheckout: Bool {
+        return additionalPaywallFields?["allowCaliforniaWebCheckout"].bool ?? false
+    }
+
     /// Local file path for the bundle - converts extractedBundleUrl to local path
     var localBundlePath: String? {
         guard let bundleUrl = extractedBundleUrl else { return nil }
--- a/Sources/Helium/HeliumCore/StripeCheckout/PaddleCheckoutPrefetchCoordinator.swift
+++ b/Sources/Helium/HeliumCore/StripeCheckout/PaddleCheckoutPrefetchCoordinator.swift
@@ -90,7 +90,8 @@ final class PaddleCheckoutPrefetchCoordinator {
             priceIds: priceIds,
             discountIdByPriceId: discountIdByPriceId,
             paddleClientToken: clientToken,
-            iosBundleId: Bundle.main.bundleIdentifier
+            iosBundleId: Bundle.main.bundleIdentifier,
+            allowCaliforniaCheckout: info.allowCaliforniaWebCheckout
         )
     }
 
@@ -106,7 +107,8 @@ final class PaddleCheckoutPrefetchCoordinator {
         priceIds: [String],
         discountIdByPriceId: [String: String] = [:],
         paddleClientToken: String,
-        iosBundleId: String?
+        iosBundleId: String?,
+        allowCaliforniaCheckout: Bool = false
     ) {
         let sessionId = paywallSession.sessionId
         let scope = paywallSession.observabilityScope
@@ -125,6 +127,7 @@ final class PaddleCheckoutPrefetchCoordinator {
                     discountId: discountId,
                     paddleClientToken: paddleClientToken,
                     iosBundleId: iosBundleId,
+                    allowCaliforniaCheckout: allowCaliforniaCheckout,
                     banditClient: bandit,
                     bffClient: bff,
                     scope: scope
@@ -454,6 +457,7 @@ final class PaddleCheckoutPrefetchCoordinator {
         discountId: String?,
         paddleClientToken: String,
         iosBundleId: String?,
+        allowCaliforniaCheckout: Bool,
         banditClient: HeliumPaymentAPIClient,
         bffClient: PaddleBFFClient,
         scope: PaywallObservabilityScope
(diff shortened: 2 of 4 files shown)


Sources/Helium/HeliumCore/StripeCheckout/PaddleCheckoutPrefetchCoordinator.swift
```diff
@@ -486,7 +490,13 @@ final class PaddleCheckoutPrefetchCoordinator {
...
                 iosBundleId: iosBundleId
             )
-            if let caPostal = californiaPostalCode(in: paddleResult.rawBody) {
+            // California (AB 2863) hard-block. Gated behind a server-config kill
+            // switch: when `allowCaliforniaCheckout` is on, CA buyers proceed to
+            // `.ready` and reach the consent-aware web paywall. Default is off
+            // (block) so binary-shipped old versions keep blocking until the
+            // bundler consent flow is confirmed live end-to-end. The ip_geo
+            // observability below still fires either way.
+            if !allowCaliforniaCheckout, let caPostal = californiaPostalCode(in: paddleResult.rawBody) {
```
salami, line 499: Can probably just remove this comment, code reads fine on its own

Sources/Helium/HeliumCore/Models.swift
```diff
@@ -92,7 +92,17 @@ public struct HeliumPaywallInfo: Codable {
...
     }
-    
+
+    /// Server-config kill switch for the California (AB 2863) checkout block.
+    /// When the web bundle's affirmative-consent flow is confirmed live, the
+    /// server sets this true and California buyers are allowed through to the
+    /// (consent-aware) web paywall instead of being hard-blocked upstream.
+    /// Defaults to `false` (block) so binary-shipped older SDK versions — and any
+    /// config that omits the field — keep the safety net until the server flips it.
+    var allowCaliforniaWebCheckout: Bool {
```
salami, line 102: Can probably just remove this comment, code reads fine on its own

Sources/Helium/HeliumCore/Models.swift
```diff
@@ -92,7 +92,17 @@ public struct HeliumPaywallInfo: Codable {
...
+        return additionalPaywallFields?["allowCaliforniaWebCheckout"].bool ?? false
```
salami, line 103: Why put in `additionalPaywallFields` instead of top-level on the paywall info map or even top-level in the overall response (so not per-trigger)?

salami (approved):
Good idea! Just a few comments to consider.

PR #299, opened 2026-07-14
[HEL-6146] | Fix triple tap opening the wrong paywall at checkout 

JesusRojass:
## What

Makes the paywall preview (triple-tap control panel) use the previewed version's **own** web checkout URL. Previously an app2web preview opened a *different* paywall's web page at checkout and errored on load. When no web paywall is linked, checkout now fails explicitly instead of opening the wrong page.

## Why (root cause)

`setPreviewTriggerConfig` builds the preview config by cloning the config of whichever existing trigger sorts first alphabetically, then overriding the bundle URL and product IDs. It never overrode `additionalPaywallFields["webPaywallBundleUrl"]`, which is exactly what `ExternalWebCheckoutManager.runCheckoutFlow` reads to decide which web page to open. So every preview kept the cloned trigger's checkout URL:

- Preview **renders** the selected paywall (bundle URL is overridden) — looks correct on screen.
- Checkout **opens** the cloned trigger's web paywall, enriched with the previewed version's product keys, which that page doesn't recognize → error on load.

## Fix

`/paywall-previews` now returns each version's own `webPaywallBundleUrl`, and `setPreviewTriggerConfig` sets it on the preview config (both the typed config and the `fetchedConfigJSON` mirror). When the response has no URL (`null`, no linked web paywall, or an older backend), the inherited value is cleared so checkout fails with the existing `webPaywallBundleUrlMissing` error instead of silently opening the wrong page.

## Tests

14 new tests in `PreviewTriggerConfigTests`, covering:
- inherited checkout URL is scrubbed / provided URL is used (typed config + JSON mirror)
- empty string treated as missing
- cloned trigger's own config stays untouched; re-preview clones the original trigger, not the previous preview
- bundle URL / product ID overrides and `forceShowFallback` clearing locked in
- decoding of both the new and the legacy `/paywall-previews` response shapes

---

> [!NOTE]
> **Low Risk**
> Scoped to debug/TestFlight preview flow and API decoding; checkout behavior for production triggers is unchanged aside from explicit nulling when preview has no web URL.
> 
> **Overview**
> Fixes **triple-tap paywall preview** opening the **wrong web paywall at app2web checkout**. Preview config was cloned from another trigger and only the in-app bundle URL was overridden, so checkout still used the donor trigger’s `webPaywallBundleUrl`.
> 
> The control panel now decodes each preview version’s **`webPaywallBundleUrl`** (and drops obsolete top-level `stripeProductIds` on the response model). **`setPreviewTriggerConfig`** applies that URL to `additionalPaywallFields`, or sets it to **null** when missing/empty so checkout fails with **`webPaywallBundleUrlMissing`** instead of loading another paywall. The control panel passes the version URL when launching a preview.
> 
> Test helpers can inject optional **`fetchedConfigJSON`**; **`PreviewTriggerConfigTests`** covers URL scrubbing, re-preview behavior, and new/legacy `/paywall-previews` decoding.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f54c3e495da181b7ed2ffa7066b10af4b59e9c44. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added support for web paywall bundle URLs in paywall previews.
  * Preview configurations now apply web checkout details while preserving original paywall settings.
  * Added compatibility for control panel responses with optional web paywall data.

* **Bug Fixes**
  * Prevented checkout URLs from leaking between repeated preview configurations.
  * Ensured missing or empty web paywall URLs are cleared correctly.

* **Tests**
  * Added coverage for preview configuration, JSON synchronization, cloning behavior, and legacy response compatibility.

### DIFF
--- a/Sources/Helium/HeliumCore/HeliumFetchedConfig.swift
+++ b/Sources/Helium/HeliumCore/HeliumFetchedConfig.swift
@@ -1092,13 +1092,6 @@ public class HeliumFetchedConfigManager {
         if var configJSON = fetchedConfigJSON {
             var sourceJSON = configJSON["triggerToPaywalls"][sourceTrigger]
             sourceJSON["resolvedConfig"]["baseStack"]["componentProps"]["bundleURL"] = JSON(bundleUrl)
-            if sourceJSON["additionalPaywallFields"].exists() {
-                if let webPaywallBundleUrl, !webPaywallBundleUrl.isEmpty {
-                    sourceJSON["additionalPaywallFields"]["webPaywallBundleUrl"] = JSON(webPaywallBundleUrl)
-                } else {
-                    sourceJSON["additionalPaywallFields"]["webPaywallBundleUrl"] = JSON.null
-                }
-            }
             configJSON["triggerToPaywalls"][Self.HELIUM_PREVIEW_TRIGGER] = sourceJSON
             fetchedConfigJSON = configJSON
         }
--- a/Sources/Helium/HeliumCore/HeliumFetchedConfig.swift
+++ b/Sources/Helium/HeliumCore/HeliumFetchedConfig.swift
@@ -214,8 +214,9 @@ public class HeliumFetchedConfigManager {
     }
 
     /// Inject config for testing purposes only. Accessible via @testable import.
-    func injectConfigForTesting(_ config: HeliumFetchedConfig) {
+    func injectConfigForTesting(_ config: HeliumFetchedConfig, json: JSON? = nil) {
         self.fetchedConfig = config
+        self.fetchedConfigJSON = json
         self._downloadStatus = .downloadSuccess
     }
 
@@ -1027,6 +1028,7 @@ public class HeliumFetchedConfigManager {
         productIdsStripe: [String],
         productIdsPaddle: [String],
         productIdsPaddleWeb: [String],
+        webPaywallBundleUrl: String? = nil,
     ) throws {
         guard var config = fetchedConfig else {
             throw HeliumControlPanelError.noConfigAvailable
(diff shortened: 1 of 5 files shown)


Sources/Helium/HeliumCore/HeliumFetchedConfig.swift
```diff
@@ -1081,6 +1092,13 @@ public class HeliumFetchedConfigManager {
         if var configJSON = fetchedConfigJSON {
             var sourceJSON = configJSON["triggerToPaywalls"][sourceTrigger]
             sourceJSON["resolvedConfig"]["baseStack"]["componentProps"]["bundleURL"] = JSON(bundleUrl)
+            if sourceJSON["additionalPaywallFields"].exists() {
+                if let webPaywallBundleUrl, !webPaywallBundleUrl.isEmpty {
+                    sourceJSON["additionalPaywallFields"]["webPaywallBundleUrl"] = JSON(webPaywallBundleUrl)
+                } else {
+                    sourceJSON["additionalPaywallFields"]["webPaywallBundleUrl"] = JSON.null
+                }
+            }
```
salami: Is this block necessary? If not let's keep it out. I think we only ever read off the above `additionalFields["webPaywallBundleUrl"]`

JesusRojass: True, upon checking all the reads nothing consumes additionalPaywallFields there.

Dropping it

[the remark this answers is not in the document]
salami: Nah, can probably strip it out. Look at my comment instead.

salami (approved):

PR #301, opened 2026-07-14
[HEL-6201] Remove HeliumPaywallEvent, add HeliumAnalyticsMapper

JesusRojass:
## What was needed

The old `HeliumPaywallEvent` enum is fully replaced by the events in PaywallEvents.swift, except for one critical use: `HeliumAnalyticsManager` still ran every event through `toLegacyEvent()` and used the enum's custom Codable encoding as the analytics payload (the `heliumEvent` object in `HeliumPaywallLoggedEvent`), plus `caseString()` for the `helium_*` track names and the trigger/template lookups for enrichment. Goal: delete the enum without changing what the backend receives.

## What was done

- Added `HeliumAnalyticsMapper` (same idea/naming as Android's `AnalyticsMapper`). Basically took the key/value mapping that lived in the enum's `encode(to:)` + each `toLegacyEvent()` and moved it into one file: `mapToAnalyticsPayload(_:)` builds the payload dict straight from the event, `getEventName` keeps the old analytics names (`offerSelected`, `subscriptionSucceeded`, `purchase_already_entitled`, etc.), and `getTriggerName`/`getPaywallTemplateName` replace `getTriggerIfExists()`/`getPaywallTemplateNameIfExists()`. All the quirks came along as-is: `errorDescription` key, `canonicalJoinTransactionId` duplication, JSON-stringified `customPaywallAction.params`, `userAllocated` only sending type + trigger. An unknown event type logs an error and sends a type-only payload instead of crashing.
- Pointed `HeliumAnalyticsManager` at the mapper. `HeliumPaywallLoggedEvent.heliumEvent` is now `SegmentJSON` — which is what Segment converted the Codable enum into internally anyway, so nothing changes on the tracking path, and the app2web checkout `JSONEncoder` path encodes it the same. Flush-on-critical-events is the same three events, just type-checked against the new structs.
- Deleted `HeliumPaywallEvent` (~565 lines) and all 21 `toLegacyEvent()` impls. Net −680 lines.
- Added `AnalyticsPayloadMappingTests`: golden tests with the exact expected payload for all 21 event types, plus track names, trigger/template extraction, and envelope nesting. Done TDD-style — the goldens were first run against both the legacy encoding and the new mapper in the same suite run to prove they produce identical output, then the legacy assertions were deleted along with the enum. The goldens stay as the permanent regression net for the wire format.

## Callouts

- **One intentional wire change:** `paywallsDownloadSuccess` no longer sends `configId` — the old path generated a random `UUID()` for it on every event, so it was noise. The real config id still goes out top-level as `fetchedConfigId`.
- **Public API removal:** `public enum HeliumPaywallEvent` is gone. Checked the RN/Expo/Flutter iOS bridges — they only use the string `"onHeliumPaywallEvent"` as a JS event-channel name, never the Swift type. Android never had this type.

## Testing

Done TDD-style since this touches critical analytics functionality:

1. **Wrote characterization tests first** (`AnalyticsPayloadMappingTests`): golden tests for all 21 event types with the exact expected payload hardcoded — every key, the `type` discriminator, all the quirks (`errorDescription`, `canonicalJoinTransactionId` duplication, stringified `params`, snake_case `purchase_already_entitled`, omitted nils). Plus tests for the `helium_*` track names, trigger/template extraction, and that the payload nests correctly under `heliumEvent` when the full `HeliumPaywallLoggedEvent` is encoded. Each event type has a fully-populated variant and a minimal one (nil optionals) where it matters.
2. **Ran the goldens against both implementations at once**: every golden asserted against the legacy `HeliumPaywallEvent` Codable encoding (ground truth) *and* the new mapper's output round-tripped through `SegmentJSON` + `JSONEncoder` — i.e. the actual wire path. Full suite green: 177 tests, 0 failures. That run is the proof the new mapper produces byte-identical output to the old encoding.
3. **Deleted the legacy enum**, removed the temporary legacy-side assertions, and ran the full suite again: 177 tests, 0 failures.

---

> [!NOTE]
> **Medium Risk**
> Touches the full Segment/Jitsu analytics pipeline and purchase-related payloads; risk is mitigated by golden wire-format tests, but the `configId` removal and public enum deletion could affect downstream consumers if anything depended on them.
> 
> **Overview**
> **Removes the legacy `HeliumPaywallEvent` enum and every `toLegacyEvent()` bridge**, routing analytics through a new **`HeliumAnalyticsMapper`** that builds the nested `heliumEvent` payload, legacy `helium_*` track names (`subscriptionSucceeded`, `ctaPressed`, etc.), and trigger/template enrichment used when logging.
> 
> **`HeliumAnalyticsManager`** now maps each `HeliumEvent` to **`SegmentJSON`** for `HeliumPaywallLoggedEvent.heliumEvent`, with a **`payloadMappingError` fallback** if serialization fails; critical-event **flush** logic uses the v2 struct types (`PaywallOpenEvent`, `PurchaseSucceededEvent`, `PaywallCloseEvent` + `isSecondTry`).
> 
> **`HeliumEvent`** gains required **`analyticsEventName`** / **`analyticsPayload()`** (no defaults) so new events must define wire mappings. **`AnalyticsPayloadMappingTests`** lock the full wire format for all 21 event types.
> 
> **Intentional wire change:** `paywallsDownloadSuccess` no longer includes a random **`configId`** in the nested payload (real id remains **`fetchedConfigId`** on the envelope). **`public enum HeliumPaywallEvent`** is removed from the SDK surface.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c256561472eda1d20e867f1d99dcb3cdf98bcc16. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **Improvements**
  - Standardized analytics payload mapping across paywall, lifecycle, purchase, and system events, with consistent `helium_` event naming and improved context (trigger and paywall template details).
  - Improved custom action parameter handling to include only valid JSON payloads.
  - Preserved/normalized purchase transaction identifiers for clearer attribution.

- **Bug Fixes**
  - Safer analytics payload construction with a minimal fallback when mapping/serialization fails.
  - Refined “critical” event delivery/flush behavior to better match event semantics.

- **Tests**
  - Added regression tests covering wire-format analytics payload mapping and logged event encoding.


salami:
**Note: this review was done with Claude assistance.**

This is a clean, well-tested refactor and the wire format looks faithfully preserved — the golden-test approach (dual-asserting against the legacy encoding before deleting it) is exactly right. The suggestions below are about hardening the *new* mapper's safety and ergonomics, not the correctness of the current output. Notably, **Android's `AnalyticsMapper` already does both of these**, so this is really about bringing the iOS mapper to parity.

### 1. Restore compile-time "did you map the new event?" safety

`mapToAnalyticsPayload` / `getEventName` switch over the `HeliumEvent` existential with a `default:` branch. That means a newly added event type compiles fine and silently emits a type-only payload (logged, not crashed) until someone remembers to add a case. The old `toLegacyEvent()` protocol requirement actually gave us this guarantee for free, and we lose it here.

Android doesn't have this problem: `HeliumEvent` is a sealed interface, so its `when (event)` blocks are **exhaustive with no `else`** — the compiler rejects a new event until it's handled in `getEventName`, `getTriggerName`, *and* `mapToAnalyticsPayload`.

Swift can't replicate that directly — a `switch` over a protocol existential (`case is X`) *requires* a `default`, so it's never exhaustiveness-checked. The equivalent guarantee in Swift is to move the payload onto the protocol as a requirement **with no default implementation**; then every conformer must implement it or it won't build. This pairs naturally with making `HeliumEvent` `internal` (which we want anyway) so the mapping method isn't exposed as public API. To keep the centralized, auditable single-file layout, the conformances can live as extensions grouped in `HeliumAnalyticsMapper.swift`:

```swift
internal protocol HeliumEvent {
    func analyticsPayload() -> [String: Any]   // no default impl → compiler-enforced
    var analyticsEventName: String { get }
}

// grouped in HeliumAnalyticsMapper.swift
extension PurchaseSucceededEvent {
    func analyticsPayload() -> [String: Any] { ... }
}
```

The one rule: **don't add a default protocol-extension impl** — that silently reintroduces the exact gap we're closing.

### 2. DRY the repeated context/product fields — like Android does

`triggerName` + `paywallTemplateName` repeat across every `PaywallContextEvent`, and `productKey` across every `ProductEvent`. Android factors these into `mapPaywallContextEvent(...)` and `mapProductEvent(...)`, with each event adding its specifics on top. Worth copying.

One guardrail: only factor the **provably uniform** fields. `paymentProcessor` looks like it belongs to a "product base," but its wire presence is *not* uniform — `offerSelected` / `subscriptionRestored` / `subscriptionPending` / `purchase_already_entitled` omit it. Android gets this right: `mapProductEvent` does **not** include `paymentProcessor`; only `mapPurchaseSucceededEvent` adds it explicitly. Keeping the quirks explicit (rather than buried in a shared base) is what keeps them reviewable — and the golden tests will catch any slip.

The one tradeoff: we lose the easy per-event eyeball check — you can no longer read an event's *complete* payload in a single block, and instead have to mentally merge the shared helper with the event's extras. In practice the reduced boilerplate, common fields that stay consistent by construction, and better signal-to-noise on the quirks outweigh it, and the golden tests remain the source of truth for the full payload.

### On `SegmentJSON`

Keeping `heliumEvent` as a JSON-tree type is the right call and mirrors Android (`JsonObject`). The throwing `SegmentJSON(...)` init is a plus — it validates eagerly at the mapping site, which is what enables the clean degrade path. Two notes:

- It couples the model to a vendored Segment type, so a one-line comment on the field explaining *why* would head off a future "cleanup" to `AnyCodable` that'd break that eager-validation ergonomics.
- We don't expect the `catch` to ever fire, but if it does, the current log line is invisible to us — this SDK runs in other people's apps, and the analytics event is the only channel we get back. Suggest the degrade payload carry a self-describing field so a serialization failure becomes queryable/alertable in the pipeline rather than silent:

  ```swift
  heliumEventPayload = .object([
      "type": .string(HeliumAnalyticsMapper.getEventName(event)),
      "payloadMappingError": .string("\(error)"),
  ])
  ```

  This is safe to build directly (it's a static `.object`, so it can't re-throw), keeps the real `type` so the event still aggregates normally, and needs only the one column — its mere presence signals the failure and the value carries the reason. Worth agreeing the exact key name with backend so there's an alert waiting for it.

### On the `configId` drop (confirmed safe)

The intentional `configId` drop from `paywallsDownloadSuccess` is the one real wire delta in this PR, and it's fully confirmed safe on both sides:

- **SDK side:** this `configId` was **never a real value** — the only code that ever set it was `let dummyConfigId = UUID()` in `PaywallsDownloadSuccessEvent.toLegacyEvent()` (a throwaway random UUID generated purely to satisfy the legacy enum's required field). This PR deletes that line along with the enum, the v2 event struct never had a `configId` field, and the emit site never passed one — so there's no orphaned code left behind and no SDK-side cleanup needed. (Distinct from the top-level `fetchedConfigId`, which is the real config id, still set from `getConfigId()` and still emitted — that one stays.)
- **Analytics pipeline:** confirmed we don't use the nested `configId` in our analytics pipeline at all.

No further action here — flagging only so the drop is documented.

Sources/Helium/HeliumCore/HeliumAnalyticsMapper.swift
```diff
@@ -1,185 +1,38 @@
...
 //  HeliumAnalyticsMapper.swift
 //
-//  Maps HeliumEvent instances to the analytics payload sent under the
-//  `heliumEvent` key of HeliumPaywallLoggedEvent. Mirrors Android's
-//  AnalyticsMapper. The keys and `type` discriminators here are the analytics
-//  wire format — they intentionally differ from HeliumEvent.eventName /
-//  toDictionary() (the public listener surface) and must not change without a
-//  backend migration.
+//  The analytics wire format previously produced by the legacy
+//  HeliumPaywallEvent enum's custom Codable encoding. Mirrors Android's
```
salami: nit - legacy HeliumPaywallEvent enum does not exist anymore after these changes. referencing adds comment bloat and is kinda confusing to new readers

Sources/Helium/HeliumCore/HeliumAnalyticsMapper.swift
```diff
@@ -197,3 +50,280 @@ enum HeliumAnalyticsMapper {
...
+            "paywallTemplateName": paywallName,
+        ]
+    }
+}
+
+extension ProductEvent {
+    /// Context fields plus `productKey`. `paymentProcessor` is intentionally NOT
+    /// here — its wire presence is not uniform (offerSelected, subscriptionRestored,
+    /// subscriptionPending, and purchase_already_entitled omit it), so events that
+    /// send it add it explicitly.
```
salami: more comment nits - no need to spell out why paymentProcessor is not here

salami (approved):

PR #310, opened 2026-07-27
[HEL-6388] | Observability event for fallback configuration

koyinusa:
> [!NOTE]
> **Low Risk**
> Additive telemetry and a backward-compatible optional scope on observability tracking; no paywall or purchase path changes.
> 
> **Overview**
> Adds **`fallback_paywalls_configured`** observability when a shipped fallback JSON loads successfully (non-empty bundles), carrying `generatedAt`, org id, trigger count, resolved trigger→paywall UUID map, and a separate **`defaultPaywallUUID`** for the iOS default trigger.
> 
> **`HeliumObservabilityManager.track`** now accepts an optional **`PaywallObservabilityScope`** so startup telemetry can run without paywall session fields; enrichment only adds paywall keys when scope is present, and event-supplied **`organizationId`** is preserved when no downloaded config exists. **`defaultFallbackTrigger`** is exposed as **`static`** on **`HeliumFallbackViewManager`** for consistent resolution and event wiring.
> 
> New tests lock the event wire format (Android-aligned) and optional-scope enrichment behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4e9fd70a328a31684519fb73c99b5002d8155523. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added an observability event for configured fallback paywalls, including configuration timestamps and trigger-to-paywall mappings.
  * Observability events can now be tracked without paywall-specific context (optional scope), with paywall fields omitted when not provided.
* **Bug Fixes**
  * Improved fallback trigger handling by consistently using the shared default fallback trigger when no valid configuration can be resolved.
* **Tests**
  * Added event wire-format tests for fallback paywalls configuration and coverage for optional paywall-scope enrichment behavior.

### DIFF
--- a/Sources/Helium/HeliumCore/HeliumFallbackViewManager.swift
+++ b/Sources/Helium/HeliumCore/HeliumFallbackViewManager.swift
@@ -49,13 +49,6 @@ public class HeliumFallbackViewManager {
                     loadedConfigJSON = json
                 }
 
-                HeliumObservabilityManager.shared.track(
-                    FallbackPaywallsConfigured(
-                        generatedAt: decodedConfig.generatedAt,
-                        triggerToPaywallUUID: decodedConfig.triggerToPaywalls.mapValues { $0.paywallUUID }
-                    )
-                )
-
                 if let bundles = loadedConfig?.bundles, !bundles.isEmpty {
                     HeliumAssetManager.shared.writeBundles(bundles: bundles)
                     let generatedAtDisplay = formatDateForDisplay(decodedConfig.generatedAt)
@@ -68,6 +61,15 @@ public class HeliumFallbackViewManager {
                     } else if generatedAtDisplay == invalidDateString {
                         HeliumLogger.log(.warn, category: .fallback, "👷 Your fallbacks are outdated! ⚠️ Consider updating them\nhttps://docs.tryhelium.com/guides/fallback-bundle")
                     }
+
+                    HeliumObservabilityManager.shared.track(
+                        FallbackPaywallsConfigured(
+                            generatedAt: decodedConfig.generatedAt,
+                            organizationID: decodedConfig.organizationID,
+                            triggerToPaywallUUID: decodedConfig.triggerToPaywalls.mapValues { $0.paywallUUID }
+                        ),
+                        scope: nil
+                    )
                 } else {
                     HeliumLogger.log(.error, category: .fallback, "👷 No bundles found in fallbacks file ‼️⚠️‼️")
                 }
--- a/Sources/Helium/HeliumCore/HeliumFallbackViewManager.swift
+++ b/Sources/Helium/HeliumCore/HeliumFallbackViewManager.swift
@@ -17,7 +17,7 @@ public class HeliumFallbackViewManager {
     
     // **MARK: - Properties**
     private let defaultFallbacksName = "helium-fallbacks"
-    private let defaultFallbackTrigger = "hlm_ios_default_flbk"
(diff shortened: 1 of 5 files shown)


Sources/Helium/HeliumCore/HeliumFallbackViewManager.swift
```diff
@@ -49,6 +49,13 @@ public class HeliumFallbackViewManager {
                     loadedConfigJSON = json
                 }
 
+                HeliumObservabilityManager.shared.track(
+                    FallbackPaywallsConfigured(
+                        generatedAt: decodedConfig.generatedAt,
+                        triggerToPaywallUUID: decodedConfig.triggerToPaywalls.mapValues { $0.paywallUUID }
+                    )
+                )
```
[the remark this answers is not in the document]
JesusRojass: Good bot

the event fires during initialize before the config fetch lands, so enrichment usually has no org to attach, and this event is only useful per-org. The parsed fallback bundle already has organizationID, so simplest fix is to include it in the event properties from decodedConfig at the emission site.

Sources/Helium/HeliumCore/Observability/HeliumObservabilityManager.swift
```diff
@@ -35,7 +35,7 @@ class HeliumObservabilityManager {
 
     func track(
         _ event: any HeliumObservabilityEvent,
-        scope: PaywallObservabilityScope
+        scope: PaywallObservabilityScope? = nil
```
JesusRojass: Tiny thing: with the = nil default, someone adding a paywall-scoped event later can forget scope: and it compiles fine, just silently loses trigger/session attribution. Keeping it required (callers write scope: nil) would make that a visible choice. Fine either way.

Sources/Helium/HeliumCore/HeliumFallbackViewManager.swift
```diff
@@ -49,6 +49,13 @@ public class HeliumFallbackViewManager {
                     loadedConfigJSON = json
                 }
 
+                HeliumObservabilityManager.shared.track(
```
JesusRojass: Should this fire when the file parses but has no bundles? A few lines down we log that case as an error, but the event still goes out, so dashboards would count the app as having fallbacks configured while the SDK treats the file as broken. Might be intentional, just want to confirm the two agree on what "configured" means.

JesusRojass (changes requested):
Looking good Brother

Just the comments plus the Bot's comments and we're golden

Sources/Helium/HeliumCore/HeliumFallbackViewManager.swift
koyinusa: Hmmm, I am not very sure. But this makes a lot of logical sense so I have moved the tracking to inside of the `bundle` check.

Sources/Helium/HeliumCore/Observability/HeliumObservabilityManager.swift
koyinusa: Updated

Sources/Helium/HeliumCore/HeliumFallbackViewManager.swift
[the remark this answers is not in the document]
koyinusa: Done

JesusRojass (approved):
LGTM!

PR #312, opened 2026-07-28
[HEL-6243] | onEntitled receives the relevant purchase/restore/skip event

JesusRojass:
## What

`onEntitled` on `presentPaywall` now receives a `PaywallEntitledEvent` argument identifying how the user became entitled: `.purchased`, `.restored`, `.alreadyEntitled` (purchase attempt resolved to an existing entitlement), or `.skipped` (paywall not shown due to `dontShowIfAlreadyEntitled`). The enum carries the concrete event and exposes `productId` and the underlying `event` for serialization.

## Why

`onEntitled` only signaled *that* the user was entitled, not *how*. Apps that need to branch on new purchase vs restore vs already-entitled (and see which product) had no way to know. Internally, the purchase/restore path only stored the session id when marking a session entitled, so the triggering event was discarded before the callback fired on paywall close.

## Fix

The entitlement marker now stores the event itself (`Set<String>` → `[String: PaywallEntitledEvent]`), captured where entitling events are fired and delivered when the paywall closes. The already-entitled skip path passes the same `PaywallSkippedEvent` instance it fires through analytics. A deprecated zero-arg `presentPaywall` overload keeps existing callers (including RN/Flutter wrappers) compiling with just a warning — no source break.

## Tests

New `PaywallEntitledEventTests` (10 tests): mark/consume semantics (once-only, last-wins), fireEvent-to-event mapping for all three entitling events plus a non-entitling negative case, both skip-path callbacks, enum accessors, and a compile-time check that all `presentPaywall` call shapes resolve. Full suite passes: 268 tests, 0 failures.

---

> [!NOTE]
> **Medium Risk**
> Public API change to a purchase/entitlement callback path; behavior is additive with a deprecated overload, but integrators must handle the new enum to branch correctly on purchase vs restore vs skip.
> 
> **Overview**
> **`presentPaywall`’s `onEntitled` now receives a `PaywallEntitledEvent`** so apps can tell *how* the user became entitled, not only that they did.
> 
> The new enum covers **`.purchased`**, **`.restored`**, **`.alreadyEntitled`**, and **`.skipped`** (when `dontShowIfAlreadyEntitled` skips the paywall). It exposes the underlying SDK event and **`productId`** where applicable.
> 
> Entitlement tracking changes from a per-session boolean to **`[sessionId: PaywallEntitledEvent]`**: entitling purchase events are wrapped in `fireEvent` and stored until paywall close, then delivered via **`consumeEntitledEvent`**. The already-entitled skip path invokes **`onEntitled(.skipped(...))`** immediately with the same `PaywallSkippedEvent` sent to analytics.
> 
> A **deprecated** zero-argument `onEntitled` overload preserves compile compatibility for existing callers. **`PaywallEntitledEventTests`** covers mark/consume, `fireEvent` wiring, and skip-path behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4a1e86078529c0092575cea539536832c7b46982. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Entitlement callbacks now include details about the entitlement outcome and associated product when applicable.
  * Added support for reporting skipped paywalls when the customer is already entitled.
  * Entitlement events are delivered when the paywall closes, including purchases, restores, and already-entitled outcomes.

* **Compatibility**
  * Existing no-argument entitlement callbacks remain supported through a deprecated compatibility option.

### DIFF
--- a/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
+++ b/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
@@ -304,21 +304,9 @@ class HeliumPaywallDelegateWrapper {
         }
         
         // Mark session for onEntitled callback on purchase/restore success
-        if let sessionId = overridePaywallSessionId ?? paywallSession?.sessionId {
-            let entitledEvent: PaywallEntitledEvent?
-            switch event {
-            case let purchased as PurchaseSucceededEvent:
-                entitledEvent = .purchased(purchased)
-            case let restored as PurchaseRestoredEvent:
-                entitledEvent = .restored(restored)
-            case let alreadyEntitled as PurchaseAlreadyEntitledEvent:
-                entitledEvent = .alreadyEntitled(alreadyEntitled)
-            default:
-                entitledEvent = nil
-            }
-            if let entitledEvent {
-                HeliumPaywallPresenter.shared.markSessionAsEntitled(sessionId: sessionId, event: entitledEvent)
-            }
+        if let sessionId = overridePaywallSessionId ?? paywallSession?.sessionId,
+           let entitledEvent = PaywallEntitledEvent(entitlingEvent: event) {
+            HeliumPaywallPresenter.shared.markSessionAsEntitled(sessionId: sessionId, event: entitledEvent)
         }
 
         if event is PaywallCloseEvent, let paywallSession {
--- a/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
+++ b/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
@@ -305,10 +305,19 @@ class HeliumPaywallDelegateWrapper {
         
         // Mark session for onEntitled callback on purchase/restore success
         if let sessionId = overridePaywallSessionId ?? paywallSession?.sessionId {
-            if event is PurchaseSucceededEvent ||
-                event is PurchaseRestoredEvent ||
-                event is PurchaseAlreadyEntitledEvent {
-                HeliumPaywallPresenter.shared.markSessionAsEntitled(sessionId: sessionId)
+            let entitledEvent: PaywallEntitledEvent?
+            switch event {
+            case let purchased as PurchaseSucceededEvent:
+                entitledEvent = .purchased(purchased)
+            case let restored as PurchaseRestoredEvent:
+                entitledEvent = .restored(restored)
+            case let alreadyEntitled as PurchaseAlreadyEntitledEvent:
+                entitledEvent = .alreadyEntitled(alreadyEntitled)
+            default:
+                entitledEvent = nil
+            }
+            if let entitledEvent {
+                HeliumPaywallPresenter.shared.markSessionAsEntitled(sessionId: sessionId, event: entitledEvent)
             }
         }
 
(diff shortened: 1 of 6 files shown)


Sources/Helium/HeliumPublic.swift
```diff
@@ -110,9 +112,26 @@ public class Helium {
             HeliumLogger.log(.debug, category: .ui, "Paywall skipped for trigger", metadata: ["trigger": trigger])
             return
         }
-        
+
         HeliumPaywallPresenter.shared.presentUpsellWithLoadingBudget(trigger: trigger, presentationContext: presentationContext)
     }
+
+    @available(*, deprecated, message: "Use onEntitled: ((PaywallEntitledEvent) -> Void) to receive the relevant purchase/restore/skip event.")
```
salami, line 119: Hmm I'm tempted to say we should just force change on app integrators, but the deprecation probably worth doing

Sources/Helium/HeliumPaywallPresentation/HeliumPaywallPresenter.swift
```diff
@@ -30,31 +30,42 @@ class HeliumPaywallPresenter {
     }
     
     /// Mark a session as having achieved entitlement (purchase/restore succeeded).
-    /// The onEntitled callback will be called when the paywall closes.
-    func markSessionAsEntitled(sessionId: String) {
-        _sessionsWithEntitlement.withValue { $0.insert(sessionId) }
+    /// The onEntitled callback will be called with the event when the paywall closes.
+    /// If a session produces multiple entitling events, the most recent one wins.
```
salami: nit - The comment implies that this might be normal behavior, but multiple entitling events are not expected and likely would indicate an underlying bug.

Sources/Helium/HeliumPaywallPresentation/HeliumPaywallPresenter.swift
```diff
@@ -30,31 +30,42 @@ class HeliumPaywallPresenter {
...
-                    paywallSession: nil
-                )
+                handleAlreadyEntitledSkip(trigger: trigger, context: context)
                 return true
             }
         }
         return false
     }
+
+    func handleAlreadyEntitledSkip(trigger: String, context: PaywallPresentationContext) {
```
salami, line 55: could this be private or do tests need it non-private?

Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
```diff
@@ -305,10 +305,19 @@ class HeliumPaywallDelegateWrapper {
...
+            switch event {
+            case let purchased as PurchaseSucceededEvent:
+                entitledEvent = .purchased(purchased)
+            case let restored as PurchaseRestoredEvent:
+                entitledEvent = .restored(restored)
+            case let alreadyEntitled as PurchaseAlreadyEntitledEvent:
+                entitledEvent = .alreadyEntitled(alreadyEntitled)
+            default:
+                entitledEvent = nil
+            }
```
salami: this switch logic could be a helper on PaywallEntitledEvent

salami (approved):
Awesome! Added some nits, feel free to address if you agree.

Sources/Helium/HeliumPaywallPresentation/HeliumPaywallPresenter.swift
JesusRojass: Tests need it non-private since it's the seam PaywallEntitledEventTests uses to exercise the skip-path callbacks, since the entitlements check itself hits StoreKit and can't be mocked.

PR #313, opened 2026-07-28
Stop routing unregistered Paddle/Stripe price keys to StoreKit

Arhan13:
## Problem

`HeliumPaymentProcessor.resolve` returns `.appStore` for any key absent from both server price maps, and `handlePurchase` hands that straight to the StoreKit delegate:

```swift
static func resolve(for productKey: String) -> HeliumPaymentProcessor {
    if stripeProducts[productKey] != nil { return .stripe }
    if paddleProducts[productKey] != nil { return .paddle }
    return .appStore
}
```

For a bare StoreKit identifier that is correct. For a `<productId>:<priceId>` composite it is not. App Store Connect has no product under `pro_01kry…:pri_01kxv…`, so `Product.products(for:)` misses and the customer gets:

> Could not find product ... in App Store Connect

That error names the wrong system. The real cause is that the paywall rang up a price the on-launch response never registered in `paddleProducts` / `stripeProducts`, which is a server or paywall configuration problem. This is not hypothetical: it is how a recent Paddle incident was initially diagnosed, and it cost real time before anyone suspected the server.

## Change

Detect the composite shape at the dispatch site and fail with an error pointing at the on-launch product list instead of App Store Connect.

The purchase failed either way. This only changes which error the developer reads, so there is no behavioural risk to a working flow.

## Why both halves are checked

`isWebProcessorCompositeKey` requires the product **and** price side to match one provider's prefix pair:

- Paddle `pro_…:pri_…`
- Stripe `prod_…:price_…`

A StoreKit identifier that merely contains a colon, or matches only one half, still routes to StoreKit exactly as before. Note `"prod_x".hasPrefix("pro_")` is false, so the two providers cannot cross-match.

## Tests

`PaymentProcessorRoutingTests` covers both provider shapes, bare and colon-containing StoreKit identifiers, an Android-style `sub:base-plan` key, half-matches, trailing colon, multiple colons, empty string, and that the error message names the product and points at the on-launch list.

## Verification

Not compiled locally. This repo is iOS-only and the machine has Command Line Tools without full Xcode, so `swift build` fails on `no such module 'UIKit'` and there is no simulator runtime. CI (`run-tests.yml`, SwiftPM tests on an iOS Simulator with Thread Sanitizer) is the first real compile. Symbols were checked by hand: `HeliumLogger` is a public enum already used in this file, `Foundation` is imported in `PaywallEvents.swift`, and `.failed` takes any `Error`.

## Related

Server-side counterpart is cloudcaptainai/bandit-server-lite-phoenix#118, which stops the two bugs that produced these unregistered keys in the first place. This PR is defence in depth for the next time it happens.

Generated with Claude Code.

---

> [!NOTE]
> **Low Risk**
> Narrow guard on misconfigured web keys only; successful StoreKit, Stripe, and Paddle paths are unchanged, with unit tests for edge cases.
> 
> **Overview**
> When a Paddle or Stripe `productId:priceId` key is missing from the on-launch price maps, `HeliumPaymentProcessor.resolve` still returns `.appStore`, so purchases used to hit StoreKit and surface misleading “not in App Store Connect” errors.
> 
> This PR adds **`isWebProcessorCompositeKey`** (Paddle `pro_…:pri_…` or Stripe `prod_…:price_…`, both halves required) and a **`HeliumPaymentRoutingError.unregisteredWebProductKey`** message that points at paywall/dashboard configuration. In **`handlePurchase`**, if routing is `.appStore` but the key matches that composite shape, the flow fails immediately with that error instead of calling **`makePurchase`**. Bare StoreKit IDs and colon-containing keys that are not web composites still go to StoreKit unchanged.
> 
> **`PaymentProcessorRoutingTests`** covers recognition, false positives, and the error copy.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 25d7376e5ee23697bc0ce1e9bf8ac1e81faf1758. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid Paddle/Stripe “web processor” composite product keys (including App Store web processor keys) now stop the purchase flow and return a clear routing failure instead of falling through.

* **Error Handling**
  * Added detection for composite key formats (`<productId>:<priceId>`) and improved the failure message when the key isn’t found in the loaded paywall products.

* **Tests**
  * Added routing tests covering valid, malformed, and unsupported key formats, plus validation of the error message content.

### DIFF
--- a/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
+++ b/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
@@ -102,6 +102,16 @@ class HeliumPaywallDelegateWrapper {
                 } catch {
                     transactionStatus = .failed(error)
                 }
+            } else if HeliumPaymentProcessor.isWebProcessorCompositeKey(productKey) {
+                // Neither price map claimed a Paddle/Stripe-shaped key, so the
+                // paywall is offering a product the on-launch response never
+                // registered. StoreKit cannot own this identifier; passing it
+                // through would surface as "product not found in App Store
+                // Connect" and send the reader hunting through App Store
+                // Connect for a product that was never meant to live there.
+                HeliumLogger.log(.error, category: .core,
+                    "Product \(productKey) is a Paddle/Stripe price key but is absent from the on-launch paddleProducts/stripeProducts maps; refusing to route it to StoreKit.")
+                transactionStatus = .failed(HeliumPaymentRoutingError.unregisteredWebProductKey(productKey))
             } else {
                 transactionStatus = await delegate.makePurchase(productId: productKey)
             }
--- a/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
+++ b/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
@@ -104,20 +104,9 @@ class HeliumPaywallDelegateWrapper {
                 }
             } else if paymentProcessor == .appStore,
                       HeliumPaymentProcessor.isWebProcessorCompositeKey(productKey) {
-                // `.appStore` means neither price map claimed the key, yet it is
-                // shaped like a Paddle/Stripe composite: the paywall is offering
-                // a product the on-launch response never registered. StoreKit
-                // cannot own this identifier, so passing it through surfaces as
-                // "product not found in App Store Connect" and sends the reader
-                // hunting through App Store Connect for a product that was never
-                // meant to live there.
-                //
-                // The processor check is load-bearing. A REGISTERED Stripe key
-                // reaches this point whenever Apple Pay is available, because
-                // the first branch requires !stripeApplePayFlowEnabled, and it
-                // has to continue on to the delegate.
-                HeliumLogger.log(.error, category: .core,
-                    "Product \(productKey) is a Paddle/Stripe price key but is absent from the on-launch paddleProducts/stripeProducts maps; refusing to route it to StoreKit.")
+                // The .appStore check is load-bearing: a registered Stripe key
+                // reaches here whenever Apple Pay is available, since the first
+                // branch requires !stripeApplePayFlowEnabled.
                 transactionStatus = .failed(HeliumPaymentRoutingError.unregisteredWebProductKey(productKey))
             } else {
                 transactionStatus = await delegate.makePurchase(productId: productKey)
--- a/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
+++ b/Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
@@ -102,13 +102,20 @@ class HeliumPaywallDelegateWrapper {
                 } catch {
                     transactionStatus = .failed(error)
                 }
-            } else if HeliumPaymentProcessor.isWebProcessorCompositeKey(productKey) {
-                // Neither price map claimed a Paddle/Stripe-shaped key, so the
-                // paywall is offering a product the on-launch response never
-                // registered. StoreKit cannot own this identifier; passing it
-                // through would surface as "product not found in App Store
-                // Connect" and send the reader hunting through App Store
-                // Connect for a product that was never meant to live there.
+            } else if paymentProcessor == .appStore,
+                      HeliumPaymentProcessor.isWebProcessorCompositeKey(productKey) {
+                // `.appStore` means neither price map claimed the key, yet it is
+                // shaped like a Paddle/Stripe composite: the paywall is offering
+                // a product the on-launch response never registered. StoreKit
+                // cannot own this identifier, so passing it through surfaces as
+                // "product not found in App Store Connect" and sends the reader
+                // hunting through App Store Connect for a product that was never
+                // meant to live there.
+                //
+                // The processor check is load-bearing. A REGISTERED Stripe key
+                // reaches this point whenever Apple Pay is available, because
+                // the first branch requires !stripeApplePayFlowEnabled, and it
+                // has to continue on to the delegate.
                 HeliumLogger.log(.error, category: .core,
                     "Product \(productKey) is a Paddle/Stripe price key but is absent from the on-launch paddleProducts/stripeProducts maps; refusing to route it to StoreKit.")
                 transactionStatus = .failed(HeliumPaymentRoutingError.unregisteredWebProductKey(productKey))
--- a/Sources/Helium/HeliumCore/PaywallEvents.swift
+++ b/Sources/Helium/HeliumCore/PaywallEvents.swift
@@ -536,6 +536,41 @@ public enum HeliumPaymentProcessor: String, Codable, Sendable {
         }
         return .appStore
     }
+
+    /// True for a `<productId>:<priceId>` composite in Paddle (`pro_…:pri_…`) or
+    /// Stripe (`prod_…:price_…`) form.
+    ///
+    /// `resolve` answers `.appStore` for anything absent from both server price
+    /// maps, which is right for a bare StoreKit identifier but not for one of
+    /// these: App Store Connect has no product under a composite key, so the
+    /// purchase can only fail, and it fails reporting a missing App Store
+    /// product rather than the missing server-side registration that actually
+    /// caused it. Both halves are checked so a StoreKit identifier that merely
+    /// happens to contain a colon isn't caught.
+    static func isWebProcessorCompositeKey(_ productKey: String) -> Bool {
+        let parts = productKey.split(separator: ":", omittingEmptySubsequences: false)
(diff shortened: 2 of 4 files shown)


Arhan13:
All three review findings addressed in `fc0c3d3`.

### Stripe Apple Pay path blocked (high severity) — real, fixed

Both bots caught this and both were right. I tested the key's *shape* without checking where it had been *routed*:

```swift
if paymentProcessor == .stripe && !stripeApplePayFlowEnabled {   // skipped when Apple Pay is on
} else if paymentProcessor == .paddle {
} else if HeliumPaymentProcessor.isWebProcessorCompositeKey(productKey) {   // caught registered Stripe keys
```

With Apple Pay available the first branch is deliberately skipped so the purchase reaches the delegate. A registered `prod_…:price_…` key then hit the shape check and was failed as unregistered, despite `resolve` having already matched it in the on-launch Stripe map. That breaks the Stripe Apple Pay flow outright, which is a worse bug than the one this PR set out to fix.

Now gated on `paymentProcessor == .appStore`, which is precisely what the branch means: neither price map claimed the key. I left a comment marking the processor check as load-bearing, since it looks redundant next to the shape check and someone will otherwise be tempted to drop it.

On the requested regression test: `handlePurchase` reaches through `HeliumFetchedConfigManager.shared`, `ApplePayHelper.shared`, and `Helium.config.purchaseDelegate`, so pinning this would mean introducing seams for all three. I have not done that here. The branch condition is now a direct restatement of the invariant and `PaymentProcessorRoutingTests` covers the predicate itself, but I would rather flag the gap than imply coverage that does not exist. Happy to add the seams in a follow-up if that is worth the churn.

### Public visibility in HeliumCore — taken

`HeliumPaymentRoutingError` and its `errorDescription` are now internal. Host apps still read the text through `localizedDescription` on the `Error` delivered by `.failed`, so the only thing lost is pattern-matching the case, which nothing needs today. Noting for the record that `StoreKitDelegateError` next door is `public` for the same role, so the convention is not applied uniformly, but reducing new surface is the cheaper default.

### Negative assertion on the diagnostic — taken, and it was a real hole

The test only required `on-launch` to appear, so a message containing both the correct guidance and the misleading destination would have passed. That is exactly the failure this PR exists to prevent, so the assertion was worth adding rather than waving off.

CI is green: `unit-tests`, `ui-tests`, and Bugbot re-review all pass on `fc0c3d3`.

Generated with Claude Code.

salami:
**Review notes**

The routing logic itself looks correct — the `paymentProcessor == .appStore` guard properly excludes registered Stripe keys (which resolve to `.stripe`), and the both-halves prefix check is a sound heuristic. A few things to tighten before merge:

**1. Trim the comments.** The 12-line block at the dispatch site, the doc comment on `isWebProcessorCompositeKey`, and the doc comment on the error case explain the same reasoning three times. Per our commenting guidance, most of this restates what the code and error message already say. The one subtlety worth keeping is a single line noting that the `.appStore` guard is what keeps registered Stripe keys (Apple Pay enabled) out of this branch — the rest can go.

**2. Move `HeliumPaymentRoutingError` out of `PaywallEvents.swift`.** That file is for event structs; the error currently sits between the `HeliumPaymentProcessor` enum and `PurchaseSucceededEvent`. `StripeCheckout/WebCheckoutModels.swift` already groups the payment/web-checkout errors (`HeliumPaymentAPIError`, `WebCheckoutError`, `PaddlePrefetchError`) and is the natural home. The `isWebProcessorCompositeKey` helper can stay on the enum.

**3. Drop the explicit `HeliumLogger.log(.error, …)` in the new branch.** Setting `.failed(HeliumPaymentRoutingError…)` fires `PurchaseFailedEvent`, and `fireEvent` already logs the event with the error's `localizedDescription` immediately after — so we emit nearly the same message twice.

**4. Reword the error's `errorDescription` for a developer audience, and loosen its test.** This string is what the integrating developer reads via `localizedDescription`; "was not in the on-launch product list" is Helium-internal phrasing they won't recognize. The "check that this product is configured on this paywall in the Helium dashboard" sentence is already the right register — lead with that framing. Note the reword is pinned by `testUnregisteredWebProductKeyErrorNamesTheProductAndTheRealCause`, which asserts the message `contains("on-launch")` and the exact `"App Store Connect"` copy; that test should assert on the product key being present rather than the prose, so the message can be edited freely.

**5. Consider (follow-up, not this PR): route to the web checkout instead of erroring.** The composite prefix already identifies the intended provider, so we could send the user to the Paddle/Stripe web checkout rather than failing. What that path actually needs is the provider's web product IDs present in the on-launch response (e.g. the `paddleProducts` map) — not the specific composite key that's routing here. That web-product registration being absent on launch is the same server/config gap that lands us in this branch, so a web route is only viable once those IDs are registered. Worth exploring as its own change, with this defense-in-depth error path as the stopgap until then.

---
_Generated by [Claude Code](https://claude.ai/code)_

Arhan13:
@salami all four actionable notes are in `25d7376`, with two places I deviated flagged below.

**1. Comments trimmed.** You're right, and it was against our own commenting guidance rather than a matter of taste: the dispatch site, the predicate doc, and the error case each restated the same reasoning. The two restatements are gone. What survives is the single line you identified as worth keeping:

```swift
// The .appStore check is load-bearing: a registered Stripe key
// reaches here whenever Apple Pay is available, since the first
// branch requires !stripeApplePayFlowEnabled.
```

That one earns its place precisely because the check reads as redundant next to the shape check, which is how the original version of this branch shipped the Apple Pay regression.

**2. Error type relocated.** Moved to `StripeCheckout/WebCheckoutModels.swift` alongside `HeliumPaymentAPIError`, `PaddlePrefetchError` and `WebCheckoutError`. `WebCheckoutError` is internal too, so the visibility matches. `isWebProcessorCompositeKey` stays on `HeliumPaymentProcessor`, which is what it describes.

**3. Duplicate log dropped.** Confirmed at `HeliumPaywallDelegate.swift:292`:

```swift
if let purchaseFailError = (event as? PurchaseFailedEvent)?.error {
    metadata["error"] = purchaseFailError.localizedDescription
}
```

One nuance worth naming rather than glossing: mine logged at `.error`, the event logs at `.info`, so at an error-only log level this diagnostic no longer appears. I don't think that justifies emitting the same text twice, since the developer gets the message through the returned error regardless of log level, but say the word if you'd rather keep the severity.

**4. Reworded, and the test loosened.** The old copy said "was not in the on-launch product list", which is vocabulary an integrating developer has never encountered. Now:

> `pro_…:pri_…` is not set up as a purchasable product on this paywall. Add it to the paywall's products in the Helium dashboard and republish, or remove it from the paywall.

I dropped the `contains("on-launch")` assertion as you asked, since pinning prose blocks future edits.

**I kept the negative assertion on `"App Store Connect"`**, which is the one deviation. It doesn't pin wording, it pins a property that should hold however the copy evolves: this message must never point at App Store Connect. That misdirection is the entire reason the error exists, and it is what sent the original Paddle investigation the wrong way. Happy to drop it if you disagree, but it felt different in kind from the prose assertion.

**5. Web-checkout routing.** Agreed, and your framing is the useful part: the composite prefix identifies the intended provider, but that path needs the provider's product IDs present in the on-launch response, and their absence is the same server gap that lands you in this branch. So the web route only becomes viable once registration is fixed, which makes this error the right stopgap rather than a placeholder for it. Not this PR.

Net: 45 non-test lines down to 32. CI green on `25d7376` (unit-tests, ui-tests, Bugbot, CodeRabbit).

Generated with Claude Code.

salami (approved):
Thanks for the quick turnaround — all four review notes are resolved:

1. **Comments trimmed.** The dispatch-site block is down to the one load-bearing fact (the `.appStore` guard keeping registered Stripe keys out when Apple Pay is enabled), and the predicate/error docs are no longer restating it.
2. **Error type relocated.** `HeliumPaymentRoutingError` now lives in `WebCheckoutModels.swift` next to the other web-checkout errors; `isWebProcessorCompositeKey` correctly stays on the enum.
3. **Double log gone.** The branch just sets `.failed(…)` and lets `fireEvent` log it once via the `PurchaseFailedEvent` metadata.
4. **Developer-facing copy.** The new message drops the internal "on-launch product list" phrasing for a concrete dashboard action, and the test now asserts on the product key + the `App Store Connect` guard rather than the exact prose — nicely decoupled.

The web-checkout routing idea remains a sensible follow-up rather than part of this defence-in-depth PR. Logic looks correct and the edge-case coverage is solid. Approving — just waiting on CI (first real compile) before merge.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #314, opened 2026-07-28
Derive web-checkout intro-offer eligibility from any offered price, not all

Arhan13:
## Problem

`isIntroOfferEligibleForWebCheckout` produces the boolean the app2web handoff writes into `ctx.introOfferEligible`, which the web bundle reads as `__heliumPaddleIntroOfferEligible` and uses to decide whether to select each bucket's trial slot.

When `/check-entitlement` returns no eligibility value, it was derived from the product set with `allSatisfy` over `subscription?.introOfferEligible`. That boolean folds two facts into one, server-side:

```go
if price.TrialPeriod != nil && price.TrialPeriod.Frequency > 0 && introEligible {
    subscription.IntroOfferEligible = true
}
```

So a price with no trial reports `false` for every customer, always, and one such product vetoed the whole signal. A web paywall offering a monthly plan alongside a yearly trial reported **every customer ineligible, permanently**, and the bundle showed no trial. Wayk's web paywalls (org `b91f1d9d`) all have that shape: every one includes the monthly no-trial bucket `0a6a685f`.

## Change

Follow the contract CML already ships as `computeIntroOfferEligibleForAll` (content-manager-lite#1170): only offer-bearing products count, every one of them must be eligible, and none offer-bearing means false.

```swift
let offerBearing = products
    .compactMap { priceMap[$0]?.subscription }
    .filter { !($0.introOffers ?? []).isEmpty }
guard !offerBearing.isEmpty else { return false }
return offerBearing.allSatisfy { $0.introOfferEligible == true }
```

Offer-bearing keys on `introOffers`, not on `introOfferEligible`. That field is not overloaded: bandit emits it only alongside `introOfferEligible: true`, and omits it both for prices with no trial and for trials the customer has consumed.

| situation | before | after | correct |
|---|---|---|---|
| eligible, all products have trials | true | true | true |
| eligible, one product has no trial | **false** | true | true |
| eligible, one product is one-time | **false** | true | true |
| customer consumed their intro offer | false | false | false |
| no product has a trial | false | false | false |

## Why not `introOfferEligible != false`

Suggested in review, and it does not fix the case. The Go field is a non-pointer bool with no `omitempty`, so a no-trial **subscription** price serialises an explicit `false` rather than omitting it. The committed golden shows `pri_paired_main` with `"introOfferEligible": false` and no `introOffers`. `!= false` only skips when the whole `subscription` block is nil, which happens for one-time products, so the monthly-plan case survives.

## Same code for both providers

Stripe eligibility is per product (`!eligibility.ProductsWithUsedTrial[productID]`), unlike Paddle's single global bit, so two trial-bearing Stripe prices can legitimately disagree. An earlier revision of this PR used `contains`, which was unsafe there. Under the offer-bearing contract a product the customer already trialed arrives with no `introOffers` and is skipped, exactly as CML treats it, so no provider branch is needed.

## Scope, and what is not established

This affects only the path taken when `entitlementsSource.introOfferEligible()` returns nil, which happens when the entitlements fetch produced no snapshot or the server omitted the field. The user-visible consequence is a missing trial on the web paywall, not a purchase failure.

**How often that path is taken has not been measured.** One case worth checking before sizing this: an unidentified user on first launch, before any `/check-entitlement` result exists. That is precisely the population that should be seeing trials.

This does not touch in-app paywall eligibility, which comes per-product from `paddleProducts`. It is also independent of cloudcaptainai/bandit-server-lite-phoenix#118, which changes only the in-app offered lists and leaves `webProductsOfferedPaddle` byte-identical.

## Tests

`WebCheckoutIntroOfferEligibilityTests` decodes fixtures from the on-launch wire shape rather than hand-constructing them, covering all five shapes: eligible trial, consumed trial, no trial, one-time, and offer-bearing-but-ineligible. The last is not something bandit emits today; it pins the contract as "every offer-bearing product is eligible" so it cannot quietly degrade to "any product is eligible" if that changes.

Generated with Claude Code.

---

> [!NOTE]
> **Medium Risk**
> Changes paywall trial signaling on the entitlements fallback path, which can affect conversion, but server eligibility is unchanged and behavior is covered by new unit tests.
> 
> **Overview**
> Fixes **web checkout** reporting every customer as intro-offer ineligible when the paywall mixes a trial plan with a **no-trial monthly** (or one-time) product, but only on the **fallback** path where `/check-entitlement` does not supply `introOfferEligible`.
> 
> The local price-map logic no longer requires `introOfferEligible == true` on **every** offered product. It now matches the CML contract via new **`blanketIntroOfferEligibility`**: consider only subscriptions with non-empty **`introOffers`**, require all of those to be eligible, and return false when none are offer-bearing. That value still flows into checkout **`ctx.introOfferEligible`** for the web bundle to pick trial slots.
> 
> Adds **`WebCheckoutIntroOfferEligibilityTests`** with JSON-decoded fixtures for the defect case, consumed trials, one-time products, and mixed eligibility.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9b1cad5c9c30d2d7774c702ad5a2a86a98094d1d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved intro-offer eligibility handling during web checkout.
  * Eligibility now correctly evaluates products that include intro offers, including mixed product types, missing data, consumed offers, and unknown products.
  * Correctly reports ineligibility when no offered product has an available intro offer.

* **Tests**
  * Added comprehensive coverage for eligible, ineligible, absent, and one-time intro-offer scenarios.

### DIFF
--- a/Sources/Helium/HeliumCore/StripeCheckout/ExternalWebCheckoutManager.swift
+++ b/Sources/Helium/HeliumCore/StripeCheckout/ExternalWebCheckoutManager.swift
@@ -183,12 +183,10 @@ public class ExternalWebCheckoutManager: NSObject {
         return try await openEnrichedCheckoutURL(enrichedURL, productKey: productKey, paywallSession: paywallSession)
     }
 
-    /// Whether this customer is intro-offer eligible, for the web bundle to
-    /// decide whether to select each bucket's trial slot.
+    /// Whether this customer is deemed intro-offer eligible for the web paywall.
     ///
     /// Prefers the server's per-customer signal from `/check-entitlement` when
-    /// available — the local price map can go stale (e.g., host app sets
-    /// userId after Helium initialized, so eligibility could change).
+    /// available — the local price map can go stale.
     private func isIntroOfferEligibleForWebCheckout(paywallInfo: HeliumPaywallInfo?) async -> Bool {
         guard let paywallInfo,
               let products = provider.getOfferedProducts(paywallInfo, false),
@@ -198,27 +196,21 @@ public class ExternalWebCheckoutManager: NSObject {
         if let serverValue = await entitlementsSource.introOfferEligible() {
             return serverValue
         }
-        return Self.introOfferEligible(
+        return Self.blanketIntroOfferEligibility(
             products: products,
             priceMap: provider.getProductsPriceMap() ?? [:]
         )
     }
 
-    /// Derives the per-customer intro-offer signal from the server price map,
-    /// for when `/check-entitlement` returns no eligibility value.
+    /// Derives the intro-offer signal from the price map when
+    /// `/check-entitlement` returns no eligibility value. Matches the contract
+    /// of CML's `computeIntroOfferEligibleForAll`.
     ///
-    /// Mirrors the coarse-prop contract CML's `computeIntroOfferEligibleForAll`
-    /// already ships: only offer-bearing products count, every one of them must
-    /// be eligible, and no offer-bearing products at all means false.
-    ///
-    /// Offer-bearing is keyed on `introOffers`, not on `introOfferEligible`.
-    /// That boolean folds two facts into one: the price carries a trial, AND
-    /// this customer has not consumed an intro offer. It is therefore false
-    /// both for a price with no trial and for a trial this customer cannot
-    /// have, and reading it alone cannot tell those apart. Testing every
-    /// offered product against it let a plain monthly plan sitting beside a
-    /// yearly trial report the customer permanently ineligible.
-    static func introOfferEligible(
+    /// Only offer-bearing prices count. A price's `introOfferEligible` is false
+    /// both when it carries no trial and when the customer already consumed
+    /// one, so requiring it of every offered product let a plain monthly plan
+    /// sitting beside a yearly trial report the customer ineligible.
+    static func blanketIntroOfferEligibility(
         products: [String],
         priceMap: [String: ServerProductPrice]
     ) -> Bool {
--- a/Sources/Helium/HeliumCore/StripeCheckout/ExternalWebCheckoutManager.swift
+++ b/Sources/Helium/HeliumCore/StripeCheckout/ExternalWebCheckoutManager.swift
@@ -183,7 +183,9 @@ public class ExternalWebCheckoutManager: NSObject {
         return try await openEnrichedCheckoutURL(enrichedURL, productKey: productKey, paywallSession: paywallSession)
     }
 
-    /// True if every offered product is intro-offer eligible.
+    /// Whether this customer is intro-offer eligible, for the web bundle to
+    /// decide whether to select each bucket's trial slot.
+    ///
     /// Prefers the server's per-customer signal from `/check-entitlement` when
     /// available — the local price map can go stale (e.g., host app sets
     /// userId after Helium initialized, so eligibility could change).
@@ -196,8 +198,33 @@ public class ExternalWebCheckoutManager: NSObject {
         if let serverValue = await entitlementsSource.introOfferEligible() {
             return serverValue
         }
-        let priceMap = provider.getProductsPriceMap() ?? [:]
-        return products.allSatisfy { priceMap[$0]?.subscription?.introOfferEligible == true }
+        return Self.introOfferEligible(
+            products: products,
+            priceMap: provider.getProductsPriceMap() ?? [:]
+        )
+    }
+
+    /// Fallback derivation of the per-customer intro-offer signal from the
+    /// server price map, used when `/check-entitlement` has no answer yet.
+    ///
+    /// A price's `subscription.introOfferEligible` already folds two things
+    /// together: the price carries a trial period, AND this customer has not
+    /// consumed an intro offer. A product with no trial therefore reports false
+    /// for everyone. Requiring EVERY offered product to report true let a
+    /// single no-trial product — a monthly plan sitting next to a yearly trial,
+    /// say — mask an eligible customer and suppress trials across the whole web
+    /// paywall.
+    ///
+    /// One product reporting true is sufficient. Eligibility is a property of
+    /// the customer, not of the product set, and the server applies that one
+    /// customer bit to every trial-bearing price. So a single true means the
+    /// customer is eligible, and an ineligible customer yields false on every
+    /// entry regardless of how many trials the paywall offers.
+    static func introOfferEligible(
+        products: [String],
+        priceMap: [String: ServerProductPrice]
+    ) -> Bool {
+        products.contains { priceMap[$0]?.subscription?.introOfferEligible == true }
     }
 
     /// Builds the enriched checkout URL: existing query items are preserved,
--- a/Sources/Helium/HeliumCore/StripeCheckout/ExternalWebCheckoutManager.swift
+++ b/Sources/Helium/HeliumCore/StripeCheckout/ExternalWebCheckoutManager.swift
@@ -203,8 +203,7 @@ public class ExternalWebCheckoutManager: NSObject {
     }
 
     /// Derives the intro-offer signal from the price map when
-    /// `/check-entitlement` returns no eligibility value. Matches the contract
-    /// of CML's `computeIntroOfferEligibleForAll`.
+    /// `/check-entitlement` returns no eligibility value.
     ///
     /// Only offer-bearing prices count. A price's `introOfferEligible` is false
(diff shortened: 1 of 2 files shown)


salami (commented):
I don't quite understand this one. Isn't checking for eligibility on all products consistent with how CML determines the top-level value? Also, paddle and Stripe currently will only treat a user as eligible if they haven't redeemed any trial or offer. I think the existing logic is also more consistent with that as well.

That being said, I can see the argument that if any trial reports as eligible, we should assume it was properly determined by bandit and ignore the other "incorrect" values.

Arhan13:
@salami raised this, quoting so the thread stands on its own:

> I don't quite understand this one. Isn't checking for eligibility on all products consistent with how CML determines the top-level value? Also, paddle and Stripe currently will only treat a user as eligible if they haven't redeemed any trial or offer. I think the existing logic is also more consistent with that as well.
>
> That being said, I can see the argument that if any trial reports as eligible, we should assume it was properly determined by bandit and ignore the other "incorrect" values.

## The concrete case

Four products on a web paywall: three with trials, customer eligible for all three, plus one product with no trial.

| | result | what the customer sees on web |
|---|---|---|
| `allSatisfy` (today) | **false** | no trial on any of the four, despite being eligible for three |
| `contains` (proposed) | **true** | trial on the three that have one, main price on the fourth |

Today's answer is false, and it is false purely because the fourth product exists.

## Why the no-trial product cannot cause harm

`productsTsxContent.ts:1085`:

```js
const useTrial = resolvedEligible && !!p.trial;
```

The bundle already checks **per bucket** whether that bucket has a trial slot, so a global `true` cannot put a trial on the no-trial product. It renders its main price either way.

Which means the global boolean only has one question to answer: is this customer intro-offer eligible.

## What each option computes

For Paddle, bandit sets the per-price value as `hasTrial(price) AND customerBit`, where `customerBit` is a single global value per request (`eligibility.IntroOfferEligible`, the unified no-trial-and-no-discount rule). So:

```
contains   = customerBit AND (at least one offered price has a trial)
allSatisfy = customerBit AND (every offered price has a trial)
```

`contains` equals `customerBit` whenever at least one trial exists, which is exactly when the answer can change anything. `allSatisfy` adds "every offered price has a trial", a fact about paywall configuration rather than about the customer, and one the bundle already applies locally per bucket. Applying it a second time globally is what lets one no-trial product veto every other bucket.

## On the CML consistency point

I don't think there is a top-level value to be consistent with. `resolvePaddleProductsAtRuntime` resolves per bucket, first-key-wins, and there is no `every`/`allSatisfy` in it. CML's own system prompt (`lib/prompts/system/sections/product-management-and-context.ts:22`) says:

> Gate trials/offers PER-PRODUCT on that product's `product.subscription?.introOfferEligible` ... never on a coarse "any product is eligible" flag.

That cuts against `contains`, but it cuts against `allSatisfy` equally, since both are coarse cross-product flags. What it really says is that the single-boolean shape is the weak point.

## Where you are right, and what I am changing

The unified-rule argument does not hold for Stripe, and this manager serves both providers:

```go
// Paddle, on_launch.go:1553 — one global customer bit
introEligible := eligibility == nil || !eligibility.CustomerExists || eligibility.IntroOfferEligible

// Stripe, on_launch.go:888 — per product
isTrialEligible = !eligibility.ProductsWithUsedTrial[productID]
```

Paddle cannot produce two trial-bearing prices that disagree. Stripe can: trialed Basic last year, never touched Pro. There `contains` would offer a Basic trial the customer cannot get. My blanket change was unsafe on that path, so I am narrowing it to Paddle and keeping `allSatisfy` for Stripe.

Worth noting `allSatisfy` is also wrong in that Stripe case, in the other direction: it suppresses the Pro trial the customer *is* entitled to. One boolean cannot express per-product eligibility, so keeping `allSatisfy` for Stripe preserves current behaviour rather than claiming to fix it.

## Why this cannot be fixed properly client-side

Both servers set `introOfferEligible` and `introOffers` only inside `if trial exists`, so on the wire these are identical:

- this price has no trial
- this price has a trial you are not eligible for

Both arrive as `introOfferEligible: false` with no `introOffers`. Filtering to trial-bearing prices, which is the derivation that would actually be correct for both providers, is not expressible from the client. That wants a bandit change to emit trial existence separately from eligibility. Separate ticket.

Generated with Claude Code.

Sources/Helium/HeliumCore/StripeCheckout/ExternalWebCheckoutManager.swift
```diff
@@ -196,8 +198,33 @@ public class ExternalWebCheckoutManager: NSObject {
         if let serverValue = await entitlementsSource.introOfferEligible() {
             return serverValue
         }
-        let priceMap = provider.getProductsPriceMap() ?? [:]
-        return products.allSatisfy { priceMap[$0]?.subscription?.introOfferEligible == true }
```
salami, line 200: I agree there is a bug, but maybe it should be
`return products.allSatisfy { priceMap[$0]?.subscription?.introOfferEligible != false }` ?

In other words, return true if all products with an offer are eligible. I think that fixes bug and maintains consistency with top-level introOfferEligible like with https://github.com/cloudcaptainai/content-manager-lite/pull/1170

I'm not overly set on this, perhaps the existing logic _is_ better.  I'm just trying to double check the decision here since it's not easy to make these kinds of changes on the SDK if we need to change again.

Arhan13:
@salami you're right about CML and I was wrong. Reworked in `28eba2b` to follow that contract.

### The CML point

There is a coarse top-level value and I missed it: `computeIntroOfferEligibleForAll` in `withPaywallState.ts`, from content-manager-lite#1170. I searched `productsTsxContent.ts` and concluded none existed, which was a bad search rather than an argument. Your consistency requirement stands.

And that PR fixes this same defect on your side, in as many words:

> a no-offer product must not force the flag false for eligible users

### Why `!= false` does not fix it here

It depends on a no-trial price sending `nil` rather than `false`. It sends `false`. The Go field is a non-pointer bool with no `omitempty`:

```go
IntroOfferEligible bool `json:"introOfferEligible"`
```

The committed golden, for a no-trial subscription price:

```
pro_paired_plan:pri_paired_main   introOfferEligible=false   introOffers=<absent>
```

So `!= false` evaluates false for it and `allSatisfy` still returns false. It only skips when the whole `subscription` block is nil, which happens for one-time products, so it would fix a trial-plus-lifetime paywall and nothing else. Wayk's monthly plan is a subscription, so the case that started this survives.

### What I changed it to

Your intent was right, the boolean just can't express it. `introOffers` can: bandit emits it only alongside `introOfferEligible: true`, and omits it both for prices with no trial and for trials the customer has consumed.

```swift
let offerBearing = products
    .compactMap { priceMap[$0]?.subscription }
    .filter { !($0.introOffers ?? []).isEmpty }
guard !offerBearing.isEmpty else { return false }
return offerBearing.allSatisfy { $0.introOfferEligible == true }
```

That is `computeIntroOfferEligibleForAll`, same rules: only offer-bearing products count, every one must be eligible, none offer-bearing is false.

Two consequences worth flagging:

**No provider branch needed.** I had been about to narrow this to Paddle, since Stripe's eligibility is per product (`!eligibility.ProductsWithUsedTrial[productID]`) and two Stripe trials can legitimately disagree. Under this contract a product the customer already trialed arrives with no `introOffers` and is skipped, which is how CML treats it too. One implementation, both providers.

**Better than the `contains` I had.** On today's wire they agree on every input, since every offer-bearing entry is eligible by construction. They diverge only if the server ever sends offers to ineligible customers, where this stays correct and `contains` would quietly become "any product is eligible". There's a test pinning that shape even though bandit doesn't emit it today.

I also retract the "this can't be fixed client-side, it needs a bandit wire change" line from my earlier comment. It can, and for the same reason I got the CML point wrong: I was reading the overloaded field.

Fixtures now cover all five wire shapes (eligible trial, consumed trial, no trial, one-time, offer-bearing-but-ineligible) rather than setting the boolean directly. CI green: unit-tests, ui-tests, Bugbot.

Your closing point is what drove this. Converging on the contract CML already shipped is better than either of us picking a third reading, especially for something this awkward to revise later.

Generated with Claude Code.

salami:
Note to @salami make sure `didCompletePurchase` only sets introOfferEligible to false if a trial or perhaps better, just rely on check entitlement response for updated value

PR #320, opened 2026-08-04
[HEL-6295] | Respect paywall scroll-lock setting in triple-tap previews

JesusRojass:
## What

Triple-tap paywall previews now respect the previewed paywall's scroll-lock setting instead of scrolling (or not) based on an unrelated paywall's config.

## Why (root cause)

The preview trigger is installed by cloning the alphabetically-first existing trigger's config and swapping in only the bundle URL and products. `shouldEnableScroll` lives in the cloned `componentProps`, so the preview rendered with the donor trigger's scroll setting rather than the previewed paywall's own.

## Fix

- Decode an optional per-version `shouldEnableScroll` from the paywall-previews response.
- `setPreviewTriggerConfig` sets it explicitly on the cloned `componentProps` in both the typed config and the JSON mirror, so the donor's value never leaks. When the response doesn't provide the field, the preview uses the platform default (scroll enabled) — same default the renderer applies to unspecified config.

Note: the previews endpoint doesn't return this field yet; until it does, previews are deterministic (default) rather than donor-dependent, and pick up the real setting as soon as the endpoint sends it.

## Tests

- Preview does not inherit the donor's scroll setting (typed config + JSON mirror).
- Response-provided value is applied; donor keeps its own setting.
- Response decoding with the new field, plus legacy responses without it.
- Full unit suite: 352 tests, 0 failures.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Scoped to debug/TestFlight preview trigger setup; no production paywall fetch path changes beyond optional API decoding.
> 
> **Overview**
> Triple-tap **paywall previews** now use the previewed version’s **`shouldEnableScroll`** instead of inheriting scroll-lock from the alphabetically first “donor” trigger when the preview config is cloned.
> 
> The control panel decodes optional **`shouldEnableScroll`** per preview version and passes it into **`setPreviewTriggerConfig`**, which writes **`componentProps.shouldEnableScroll`** on both the typed config and the JSON mirror (default **`true`** when the API omits the field, matching the renderer). Until the previews endpoint sends the field, behavior is deterministic scroll-on rather than donor-dependent.
> 
> Tests cover donor isolation, explicit values, JSON mirror updates, and legacy response decoding.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f159c55f35ed8f6921acc90cf19bebe5b339f515. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added configurable scrolling for paywall previews.
  - Preview versions can explicitly enable or disable scrolling independently of the source paywall.
  - When no setting is provided, preview scrolling is enabled by default.
- **Bug Fixes**
  - Preview configurations now consistently preserve their own scroll settings without inheriting unintended values from the source paywall.
  - Existing preview responses remain compatible when the new setting is absent.

### DIFF
--- a/Sources/Helium/HeliumCore/ControlPanel/HeliumControlPanelModels.swift
+++ b/Sources/Helium/HeliumCore/ControlPanel/HeliumControlPanelModels.swift
@@ -28,6 +28,7 @@ struct HeliumPaywallPreviewVersion: Codable, Identifiable {
     let webPaddleProductIds: [String]?
     let webStripeProductIds: [String]?
     let webPaywallBundleUrl: String?
+    let shouldEnableScroll: Bool?
     let lastSavedAt: String?
     var id: String { versionId }
 
--- a/Sources/Helium/HeliumCore/ControlPanel/HeliumControlPanelView.swift
+++ b/Sources/Helium/HeliumCore/ControlPanel/HeliumControlPanelView.swift
@@ -348,7 +348,8 @@ struct HeliumControlPanelView: View {
                     productIdsPaddle: version.paddleProductIds ?? [],
                     productIdsPaddleWeb: version.webPaddleProductIds ?? [],
                     productIdsStripeWeb: version.webStripeProductIds ?? [],
-                    webPaywallBundleUrl: version.webPaywallBundleUrl
+                    webPaywallBundleUrl: version.webPaywallBundleUrl,
+                    shouldEnableScroll: version.shouldEnableScroll
                 )
 
                 guard !Task.isCancelled else {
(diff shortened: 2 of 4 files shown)


Sources/Helium/HeliumCore/ControlPanel/HeliumControlPanelModels.swift
```diff
@@ -28,6 +28,7 @@ struct HeliumPaywallPreviewVersion: Codable, Identifiable {
     let webPaddleProductIds: [String]?
     let webStripeProductIds: [String]?
     let webPaywallBundleUrl: String?
+    let shouldEnableScroll: Bool?
```
salami, line 31: Is bandit sending this already?

JesusRojass: Forgot to mention I will need a follow up ticket

JesusRojass: Done and if deploy come out good we are ready for this change

salami (approved):

PR #324, opened 2026-08-06
[SDK-48] | Combine web checkout success and cancel URLs into a single redirect URL

JesusRojass:
## What

Replaces the two-URL External Web Checkout setup with a single redirect URL. New API: `enableExternalWebCheckout(redirectURL:paymentProcessors:)`. The existing `successURL:cancelURL:` overload is deprecated but keeps its exact behavior.

## Why

Customers are told to use the same URL for success and cancel anyway (SDK-48). The SDK already distinguishes outcomes itself — query params appended at redirect time (`transactionId`, `alreadySubscribed`, `error`), and entitlement reconciliation on foreground when those are inconclusive — so separate URLs only complicate setup.

## Fix

- Added `enableExternalWebCheckout(redirectURL:paymentProcessors:)`; both public overloads route through a shared private helper. Query-bearing redirect URLs are accepted with a warning log (cancel detection degrades to the foreground observer).
- `WebCheckoutRedirect.classify` logic unchanged — the identical-URL path was already supported; doc comments now describe it as the standard configuration.
- Internal storage still holds two URLs so legacy distinct-URL configs behave as before, and the web bundle keeps receiving all three ctx keys (`successUrl`, `cancelUrl`, `paymentFailureUrl`).
- Fixed the `handleURL` docstring (non-compiling example, stale API reference) and the `checkoutURLsNotConfigured` error message.
- Example app moved to the new API.

## Tests

- New `WebCheckoutRedirectTests` (19 tests): first coverage of `WebCheckoutRedirect.classify` — query-param disambiguation, bare-URL-means-cancel, inconclusive-means-nil, Stripe `{CHECKOUT_SESSION_ID}` prefix matching, slash normalization, legacy distinct-URL matching, and config validation.
- Full unit suite passes on iPhone 17 Pro simulator.

## Related

- Docs (Paddle/Stripe onboarding guides) show the two-arg call; need the one-arg call after release.
- React Native/Flutter wrappers should adopt the single-URL API and mirror the deprecation.

---

> [!NOTE]
> **Medium Risk**
> Touches public checkout configuration and deep-link handling on the purchase path, but redirect classification behavior is unchanged and the old API is preserved via deprecation.
> 
> **Overview**
> External web checkout setup is simplified to **one redirect URL** via `Helium.config.enableExternalWebCheckout(redirectURL:paymentProcessors:)`, which stores that URL for both success and cancel internally. The **two-URL overload remains** but is **deprecated**; both paths now share a private `setExternalWebCheckout` helper with the same validation as before.
> 
> **Docs and DX:** `WebCheckoutRedirect.classify` comments now describe the single-URL model as standard (query params for outcome, bare match = cancel). `handleURL` docs fix the `.onOpenURL` example (`== nil` instead of negating a non-Bool) and point at the new API. The misconfiguration error text refers to the redirect URL API. The example app switches from `successURL`/`cancelURL` to `redirectURL`.
> 
> **Tests:** New `WebCheckoutRedirectTests` cover single-URL query disambiguation, legacy distinct URLs, URL normalization, Stripe session placeholder matching, and config validation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3630cf35b53fe29b05920c2d1528c719862f4cf7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring external web checkout with a single redirect URL for successful, canceled, and failed checkout outcomes.
  * Improved redirect handling, including URL normalization and outcome classification.
  * Added clearer validation and warnings for redirect URL configurations.

* **Deprecations**
  * The separate success and cancellation URL configuration remains available but is now deprecated in favor of the unified redirect URL approach.

* **Documentation**
  * Updated checkout and URL-handling guidance to clarify redirect behavior and entitlement reconciliation.

### DIFF
--- a/Sources/Helium/HeliumPublic.swift
+++ b/Sources/Helium/HeliumPublic.swift
@@ -430,8 +430,11 @@ public class Helium {
     /// This is not required, but encouraged for smoother post-purchase experience.
     ///
     /// Safe to call with unrelated URLs — returns `nil` if external web checkout is
-    /// disabled or the URL does not match the URLs configured via
-    /// ``HeliumConfig/enableExternalWebCheckout(successURL:cancelURL:)``.
+    /// disabled, the URL does not match a redirect URL configured via
+    /// ``HeliumConfig/enableExternalWebCheckout(redirectURL:paymentProcessors:)``,
+    /// or the checkout outcome can't be determined from the URL alone. In that last
+    /// case Helium still reconciles the purchase against entitlement state when the
+    /// app returns to the foreground.
     ///
     /// Call this from `.onOpenURL`, `SceneDelegate.scene(_:openURLContexts:)`, or
     /// `AppDelegate.application(_:open:options:)`.
@@ -439,7 +442,7 @@ public class Helium {
     /// ## Example
     /// ```swift
     /// .onOpenURL { url in
-    ///     if !Helium.shared.handleURL(url) {
+    ///     if Helium.shared.handleURL(url) == nil {
     ///         // your own deep link handling
     ///     }
     /// }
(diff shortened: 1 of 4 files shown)


Sources/Helium/HeliumPublic.swift
```diff
@@ -747,13 +750,36 @@ public class HeliumConfig {
...
+    /// determines the outcome from query parameters the checkout flow appends, and otherwise
+    /// reconciles against entitlement state when the app returns to the foreground.
+    ///
+    /// Register the URL as a deep link (custom scheme or universal link) and forward it to
+    /// ``Helium/handleURL(_:)`` from your URL handler.
     ///
     /// - Parameters:
-    ///   - successURL: The URL to redirect to after a successful payment.
-    ///   - cancelURL: The URL the provider redirects to when the user cancels checkout.
+    ///   - redirectURL: The URL checkout redirects back to when the user is done, regardless of outcome.
```
salami: nit - can remove "regardless of outcome" that's not adding value especially if you have no context of the previous separation of the urls

Sources/Helium/HeliumPublic.swift
```diff
@@ -430,16 +430,19 @@ public class Helium {
...
+    /// app returns to the foreground.
     ///
     /// Call this from `.onOpenURL`, `SceneDelegate.scene(_:openURLContexts:)`, or
     /// `AppDelegate.application(_:open:options:)`.
     ///
     /// ## Example
     /// ```swift
     /// .onOpenURL { url in
-    ///     if !Helium.shared.handleURL(url) {
+    ///     if Helium.shared.handleURL(url) == nil {
```
salami, line 442: good fix

Sources/Helium/HeliumPublic.swift
```diff
@@ -747,13 +750,36 @@ public class HeliumConfig {
...
+        redirectURL: String,
+        paymentProcessors: WebCheckoutProcessors
+    ) {
+        guard let parsed = URL(string: redirectURL), parsed.scheme != nil else {
+            HeliumLogger.log(.error, category: .core, "enableExternalWebCheckout: invalid redirectURL. It must be a valid URL with a scheme (e.g. https://example.com or myapp://path).")
+            return
+        }
+        if parsed.query != nil {
+            HeliumLogger.log(.warn, category: .core, "enableExternalWebCheckout: redirectURL contains query parameters. Checkout outcome detection relies on query parameters appended at redirect time, so prefer a query-free URL.")
+        }
```
salami: In theory we should already be handling query params just fine. Not sure if this warning is needed.

Sources/Helium/HeliumPublic.swift
```diff
@@ -747,13 +750,36 @@ public class HeliumConfig {
     /// Enables External Web Checkout Flow for any Paddle or Stripe products in your paywalls. If not enabled, paywalls with Paddle/Stripe products
     /// will not show. Your fallback paywall/s, if provided, will show instead.
     ///
-    /// You must provide redirect URLs so Helium knows where to send the user after checkout completes or is cancelled.
-    /// It is ok to use the same url for both success and cancel.
+    /// You must provide a redirect URL so Helium knows where to send the user back after checkout,
+    /// whether it succeeded, was cancelled, or failed. The one URL covers all outcomes — the SDK
+    /// determines the outcome from query parameters the checkout flow appends, and otherwise
+    /// reconciles against entitlement state when the app returns to the foreground.
```
salami: Integrator does not need to know that url "covers all outcomes", that is spelled out in the previous sentence. Suggestion:
"You must provide a redirect URL so Helium knows where to send the user back after checkout, whether it succeeded, was cancelled, or failed. If a user returns to the app manually without the redirect URL, then Helium SDK will look at latest entitlement state to determine if a purchase was made."

Sources/Helium/HeliumPublic.swift
```diff
@@ -430,16 +430,19 @@ public class Helium {
     /// This is not required, but encouraged for smoother post-purchase experience.
     ///
     /// Safe to call with unrelated URLs — returns `nil` if external web checkout is
-    /// disabled or the URL does not match the URLs configured via
-    /// ``HeliumConfig/enableExternalWebCheckout(successURL:cancelURL:)``.
+    /// disabled, the URL does not match a redirect URL configured via
+    /// ``HeliumConfig/enableExternalWebCheckout(redirectURL:paymentProcessors:)``,
+    /// or the checkout outcome can't be determined from the URL alone. In that last
+    /// case Helium still reconciles the purchase against entitlement state when the
+    /// app returns to the foreground.
```
salami: can remove, this is repeated below

salami (approved):

PR #311, opened 2026-07-27
[HEL-6281] | Disable second try flows in paywall previews

JesusRojass:
## What

Disables second try flows entirely when a paywall is opened from a triple-tap paywall preview. Instead of presenting the second try paywall, the SDK shows the paywall diagnostic screen explaining that second try flows aren't supported in previews and that the paywall should be added to a workflow to test it.

## Why

Testing second try flows from triple-tap previews doesn't work, especially when the paywall isn't in a workflow yet: the uuid / `<trigger>_second_try` lookups resolve against the fetched workflow config, which the previewed paywall may not be in. For now the flow is disabled in previews outright and says so (HEL-6281).

## Fix

- `HeliumActionsDelegate.showSecondaryPaywall` — the single entry point for second try presentation (both the paywall-uuid path and the `<trigger>_second_try` path) — returns early when the current trigger is `HELIUM_PREVIEW_TRIGGER` and presents the diagnostic instead.
- New authored record `mapSecondTryInPreview()` in `DiagnosticContentMapper`: `.expected` category, "add this paywall to a workflow" remediation, "Open Workflows" CTA, reason code `secondTryNotSupportedInPreview`. It's a separate entry point rather than a new `PaywallUnavailableReason` because the paywall itself rendered fine — no not-shown event carries this state, and no `PaywallOpenFailedEvent` is fired.
- The diagnostic presents in its own window above the preview paywall, which stays on screen behind it.

## Tests

- Added the new record to the `allContent()` invariant sweep in `DiagnosticContentMapperTests` (no URLs or raw reason codes in prose, authored links openable).
- New `testSecondTryInPreviewAdvisesAWorkflowAndReassuresAboutRealUsers` pins the category, copy, CTA, and reason code.

---

> [!NOTE]
> **Low Risk**
> Scoped to preview trigger only; production second-try behavior is unchanged and the change is diagnostic UX plus an early return.
> 
> **Overview**
> Triple-tap **paywall preview** no longer attempts to open a second-try paywall. When the paywall asks for a secondary flow while the session uses `helium_preview_trigger`, `showSecondaryPaywall` returns early and shows the paywall diagnostic instead of uuid / `_second_try` lookup and presentation.
> 
[...]

PR #328, opened 2026-08-12
[HEL-6506] | Second try paywall preview support

JesusRojass:
## What

Second try flows now work from triple-tap paywall previews, replacing the blanket "not supported in previews" block from #311.

- The control panel downloads the second try bundle alongside the main one (concurrently via `async let`) and installs it as a `helium_preview_trigger_second_try` entry; a download failure never blocks the main preview, it just flips a `loadFailed` state.
- `showSecondaryPaywall` on the preview trigger presents the armed second try through the normal presentation path. uuid lookups are skipped in previews so a preview can never present a production paywall.
- The diagnostic now covers the two remaining cases with distinct copy: no second try configured (`secondTryNotConfiguredInPreview`) and bundle load failure (`secondTryPreviewLoadFailed`).
- Preview-only behaviors (no silent fallback substitution, diagnostics always allowed, triple-tap returns to the panel) apply to the second try preview via a shared `isPreviewTrigger` helper. That also closes a latent donor-selection hole where a stale synthetic entry could have been cloned as a donor.

## Server contract

Consumes the new `secondTry` field from cloudcaptainai/bandit-server-lite-phoenix#136. Against a Bandit without it, `secondTry` decodes nil and previews show the not-configured diagnostic — no breakage either way.

## Decisions worth a second opinion

- A never-published (draft-only) second try is previewable, matching the web-paywall preview precedent, but logs a "real users will not see it until the paywall is published" warning and carries `versionStatus` for future UI.
- Preview second try presentations reuse the same session/context semantics as production second tries, so event behavior is unchanged in shape.

## Testing

Full unit suite green on iPhone 17 Pro simulator. New coverage: second try trigger install/clear, donor selection skipping synthetic entries, fallback-preview arming clearing the second try, JSON mirror install/clear, state transitions, response decoding, and the two diagnostic records.

---

> [!NOTE]
> **Medium Risk**
> Touches preview config mutation, second-try presentation routing, and paywall dismiss/navigation in dev-only paths; production `showSecondaryPaywall` gains an explicit preview guard on UUID lookup but behavior is otherwise unchanged.
> 
> **Overview**
> **Second try flows now work in paywall previews** instead of blocking with a dedicated “not supported in preview” diagnostic. The control panel decodes optional `secondTry` on each preview entry, fetches that bundle concurrently with the main preview (`async let`), and installs it under `helium_preview_trigger_second_try` via `setPreviewTriggerConfig`; a failed second-try download only logs and leaves the main preview intact, with missing config surfacing the normal `secondTryNoMatch` path.
> 
> `showSecondaryPaywall` on preview triggers skips UUID→live-trigger resolution so previews cannot accidentally present production paywalls, then presents through the usual `trigger_second_try` lookup. Shared **`isPreviewTrigger`** covers both synthetic triggers so preview rules apply to the second-try layer too: no silent fallback substitution, triple-tap **steps back one layer** (second try → main preview → panel), and the control panel **activity lock** releases only when the main preview trigger closes.
> 
> **`isFallbackPreviewArmed`** moves to `HeliumControlPanelService`; fallback and dashboard preview install/clear paths also remove stale second-try synthetic entries so donor selection cannot clone a leftover preview trigger.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d717a7625550c14491b2b3ef70c850f9a760c857. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added support for optional “second-try” paywall previews, including product details, bundle configuration, and scroll settings.
  - Preview flows can present a configured second-try paywall when available.
  - Added support for fallback preview state across preview flows.

- **Bug Fixes**
  - Improved preview trigger recognition across preview and fallback experiences.
  - Primary previews continue to open when an optional second-try preview cannot be loaded.
  - Improved cleanup of outdated preview configuration during fallback and reset flows.

### DIFF
--- a/Sources/Helium/HeliumCore/Components/Web/DynamicWebView.swift
+++ b/Sources/Helium/HeliumCore/Components/Web/DynamicWebView.swift
@@ -487,7 +487,8 @@ fileprivate struct WebViewRepresentable: UIViewRepresentable {
         @objc func handleTripleTap() {
             guard HeliumControlPanelService.shared.allowPaywallControlPanel else { return }
             if HeliumFetchedConfigManager.isPreviewTrigger(trigger) {
-                // Just return to underlying control panel instead of stacking
+                // Step back one layer instead of stacking another panel: the main preview returns
+                // to the control panel, a second try preview returns to the main preview.
                 HeliumPaywallPresenter.shared.hideUpsell()
                 return
             }
--- a/Sources/Helium/HeliumCore/Components/Web/DynamicWebView.swift
+++ b/Sources/Helium/HeliumCore/Components/Web/DynamicWebView.swift
@@ -486,7 +486,7 @@ fileprivate struct WebViewRepresentable: UIViewRepresentable {
 
         @objc func handleTripleTap() {
             guard HeliumControlPanelService.shared.allowPaywallControlPanel else { return }
-            if trigger == HeliumFetchedConfigManager.HELIUM_PREVIEW_TRIGGER {
+            if HeliumFetchedConfigManager.isPreviewTrigger(trigger) {
                 // Just return to underlying control panel instead of stacking
                 HeliumPaywallPresenter.shared.hideUpsell()
                 return
--- a/Sources/Helium/HeliumCore/ControlPanel/HeliumControlPanelView.swift
+++ b/Sources/Helium/HeliumCore/ControlPanel/HeliumControlPanelView.swift
@@ -389,7 +389,9 @@ struct HeliumControlPanelView: View {
             HeliumLogger.log(.warn, category: .ui, "[HeliumControlPanel] Second try paywall has no bundle URL", metadata: [
                 "secondTryPaywall": secondTry.paywallName,
             ])
-            return (nil, true)
+            // No bundle was ever built for this second try, so the retry advice of the
+            // load-failed diagnostic cannot resolve it; the dashboard remediation can.
+            return (nil, false)
         }
         do {
             let (bundleId, html) = try await HeliumControlPanelService.shared.fetchSingleBundle(bundleURL: secondTryBundleUrl)
@@ -401,7 +403,7 @@ struct HeliumControlPanelView: View {
                 productIdsStripe: secondTry.stripeProductIds ?? [],
                 productIdsPaddle: secondTry.paddleProductIds ?? [],
                 shouldEnableScroll: secondTry.shouldEnableScroll,
-                versionStatus: secondTry.versionStatus
+                versionStatus: secondTry.versionStatus ?? "draft"
             ), false)
         } catch {
             HeliumLogger.log(.warn, category: .ui, "[HeliumControlPanel] Failed to load second try bundle for preview", metadata: [
--- a/Sources/Helium/HeliumCore/ControlPanel/HeliumControlPanelView.swift
+++ b/Sources/Helium/HeliumCore/ControlPanel/HeliumControlPanelView.swift
@@ -339,7 +339,7 @@ struct HeliumControlPanelView: View {
             do {
                 async let secondTryTask = fetchSecondTryBundle(for: paywall)
                 let (bundleId, html) = try await HeliumControlPanelService.shared.fetchSingleBundle(bundleURL: bundleUrl)
-                let secondTryResult = await secondTryTask
+                let secondTryBundle = await secondTryTask
 
                 try HeliumFetchedConfigManager.shared.setPreviewTriggerConfig(
                     bundleId: bundleId,
@@ -352,11 +352,8 @@ struct HeliumControlPanelView: View {
                     productIdsStripeWeb: version.webStripeProductIds ?? [],
                     webPaywallBundleUrl: version.webPaywallBundleUrl,
                     shouldEnableScroll: version.shouldEnableScroll,
-                    secondTry: secondTryResult.bundle
+                    secondTry: secondTryBundle
                 )
-                if secondTryResult.loadFailed {
-                    HeliumFetchedConfigManager.shared.markPreviewSecondTryLoadFailed()
-                }
 
                 guard !Task.isCancelled else {
                     await MainActor.run { activity = .idle }
@@ -377,40 +374,31 @@ struct HeliumControlPanelView: View {
     }
 
     /// Resolves a paywall's second try flow into a preview bundle. The second try must not block
-    /// or fail the main preview, so a download error only marks the load as failed — the preview
-    /// presents without a second try and the diagnostic explains why when one is requested.
+    /// or fail the main preview, so a failure only logs — the preview presents without a second
+    /// try entry and a request for it surfaces the standard second-try-no-match diagnostic.
     private func fetchSecondTryBundle(
         for paywall: HeliumPaywallPreviewEntry
-    ) async -> (bundle: HeliumFetchedConfigManager.PreviewSecondTryBundle?, loadFailed: Bool) {
-        guard let secondTry = paywall.secondTry else {
-            return (nil, false)
-        }
-        guard let secondTryBundleUrl = secondTry.bundleUrl else {
-            HeliumLogger.log(.warn, category: .ui, "[HeliumControlPanel] Second try paywall has no bundle URL", metadata: [
-                "secondTryPaywall": secondTry.paywallName,
-            ])
-            // No bundle was ever built for this second try, so the retry advice of the
-            // load-failed diagnostic cannot resolve it; the dashboard remediation can.
-            return (nil, false)
+    ) async -> HeliumFetchedConfigManager.PreviewSecondTryBundle? {
+        guard let secondTry = paywall.secondTry, let secondTryBundleUrl = secondTry.bundleUrl else {
+            return nil
         }
         do {
             let (bundleId, html) = try await HeliumControlPanelService.shared.fetchSingleBundle(bundleURL: secondTryBundleUrl)
-            return (HeliumFetchedConfigManager.PreviewSecondTryBundle(
+            return HeliumFetchedConfigManager.PreviewSecondTryBundle(
                 bundleId: bundleId,
                 bundleUrl: secondTryBundleUrl,
                 bundleHtml: html,
                 productIds: secondTry.productIds ?? [],
                 productIdsStripe: secondTry.stripeProductIds ?? [],
                 productIdsPaddle: secondTry.paddleProductIds ?? [],
-                shouldEnableScroll: secondTry.shouldEnableScroll,
-                versionStatus: secondTry.versionStatus ?? "draft"
-            ), false)
+                shouldEnableScroll: secondTry.shouldEnableScroll
+            )
         } catch {
             HeliumLogger.log(.warn, category: .ui, "[HeliumControlPanel] Failed to load second try bundle for preview", metadata: [
                 "secondTryPaywall": secondTry.paywallName,
                 "error": "\(error)",
             ])
-            return (nil, true)
+            return nil
         }
     }
 
@@ -436,7 +424,13 @@ struct HeliumControlPanelView: View {
     private func previewPresentationContext() -> PaywallPresentationContext {
         PaywallPresentationContext(
             config: PaywallPresentationConfig(dontShowIfAlreadyEntitled: false),
-            eventHandlers: PaywallEventHandlers().onClose { _ in activity = .idle },
+            // A second try preview shares this context and closes back onto the main preview,
+            // so only the main preview trigger closing releases the lock.
+            eventHandlers: PaywallEventHandlers().onClose { event in
+                if event.triggerName == HeliumFetchedConfigManager.HELIUM_PREVIEW_TRIGGER {
+                    activity = .idle
+                }
+            },
             onEntitled: nil,
             onPaywallNotShown: { _ in activity = .idle }
         )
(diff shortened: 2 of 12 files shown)


Sources/Helium/HeliumCore/Components/Web/DynamicWebView.swift
```diff
@@ -486,7 +486,7 @@ fileprivate struct WebViewRepresentable: UIViewRepresentable {
 
         @objc func handleTripleTap() {
             guard HeliumControlPanelService.shared.allowPaywallControlPanel else { return }
-            if trigger == HeliumFetchedConfigManager.HELIUM_PREVIEW_TRIGGER {
+            if HeliumFetchedConfigManager.isPreviewTrigger(trigger) {
```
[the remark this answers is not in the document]
JesusRojass: Intended behavior, but the comment was misleading so I fixed it in c45e12a. Triple-tap steps back one layer at a time: second try preview returns to the main preview, and triple-tapping that returns to the control panel. Same shape as Android's GoBack navigation, and it lets you hop between the two paywalls while testing instead of getting yanked all the way back to the panel.

Sources/Helium/HeliumCore/ControlPanel/HeliumControlPanelView.swift
```diff
@@ -370,6 +376,42 @@ struct HeliumControlPanelView: View {
...
+            ), false)
+        } catch {
+            HeliumLogger.log(.warn, category: .ui, "[HeliumControlPanel] Failed to load second try bundle for preview", metadata: [
+                "secondTryPaywall": secondTry.paywallName,
+                "error": "\(error)",
+            ])
+            return (nil, true)
+        }
+    }
+
```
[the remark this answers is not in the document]
JesusRojass: Good catch, fixed in c45e12a. The second try preview now presents with a context that only carries over the presentation config — not the panel's close/not-shown handlers — so closing the second try no longer flips the panel's activity lock while the main preview is still on screen. The panel unlocks only when the main preview itself closes.

Sources/Helium/HeliumCore/HeliumFetchedConfig.swift
```diff
@@ -1166,27 +1276,31 @@ public class HeliumFetchedConfigManager {
...
         var sourceJSON = configJSON["triggerToPaywalls"][sourceTrigger]
         sourceJSON["resolvedConfig"]["baseStack"]["componentProps"]["bundleURL"] = JSON(bundleUrl)
         sourceJSON["resolvedConfig"]["baseStack"]["componentProps"]["shouldEnableScroll"] = JSON(shouldEnableScroll ?? true)
         configJSON["triggerToPaywalls"][HELIUM_PREVIEW_TRIGGER] = sourceJSON
+
+        if let secondTry {
+            var secondTryJSON = configJSON["triggerToPaywalls"][sourceTrigger]
+            secondTryJSON["resolvedConfig"]["baseStack"]["componentProps"]["bundleURL"] = JSON(secondTry.bundleUrl)
+            secondTryJSON["resolvedConfig"]["baseStack"]["componentProps"]["shouldEnableScroll"] = JSON(secondTry.shouldEnableScroll ?? true)
+            configJSON["triggerToPaywalls"][HELIUM_PREVIEW_SECOND_TRY_TRIGGER] = secondTryJSON
```
salami: These 4 lines copy 1290-1293 can we extract into a helper?

Sources/Helium/HeliumCore/Diagnostic/Mapper/DiagnosticContentMapper.swift
```diff
@@ -113,22 +113,39 @@ struct DiagnosticContentMapper {
...
+            )
+        case .notConfigured, .armed:
+            return DiagnosticContent(
+                category: .expected,
+                title: "No second try flow to preview",
+                body: "This paywall asked Helium to open its second try paywall, but no second "
+                    + "try flow is configured for it. Configure one in the dashboard, then "
+                    + "reopen this preview to test it.",
+                usersWillSee: "Users stay on this paywall when it requests its second try, until "
+                    + "a second try flow is configured and the paywall is published.",
```
salami: The "until" is kinda confusing. Could be read as user staying on paywall until second try flow configured. I think `usersWillSee` could just be "Users will not see a second try flow."

Sources/Helium/HeliumCore/HeliumActionsDelegate.swift
```diff
@@ -200,8 +203,33 @@ class HeliumActionsDelegate: ObservableObject {
...
+                eventHandlers: nil,
+                onEntitled: nil,
+                onPaywallNotShown: nil
+            )
+            HeliumPaywallPresenter.shared.presentUpsell(trigger: secondTryTrigger, isSecondTry: true, presentationContext: secondTryContext)
+            return
+        }
+
+        // A second try preview requesting its own second try lands here too and gets the
+        // not-configured explanation rather than another level of nesting.
```
salami: Is this even possible? Not sure I understand what's being described here.

Sources/Helium/HeliumCore/HeliumFetchedConfig.swift
```diff
@@ -237,6 +238,7 @@ public class HeliumFetchedConfigManager {
     @HeliumAtomic private(set) var triggersWithSkippedBundleAndReason: [(trigger: String, reason: PaywallUnavailableReason)] = []
     @HeliumAtomic private var localizedPriceMap: [String: LocalizedPrice] = [:]
     @HeliumAtomic private(set) var isFallbackPreviewArmed: Bool = false
+    @HeliumAtomic private(set) var previewSecondTryState: PreviewSecondTryState = .notConfigured
```
salami: I'd prefer this kind of preview state not live on HeliumFetchedConfig so we have a better separation of paywall-preview logic from core paywall presentation. Could we maybe move it to HeliumControlPanelService? Same goes for isFallbackPreviewArmed.

salami (commented):
Do we even need to store `previewSecondTryState`? What if we remove that, remove `showSecondaryPaywallForPreview` altogether, remove 174-179 of HeliumActionsDelegate.swift and just use the normal second try flow presentation?

This would be truer to the real presentation flow that we are trying to emulate and simplify things. Note that if this is doable then some of my other PR comments may no longer apply.

JesusRojass:
Took the simplification in 43bc702 — you were right that it works: `previewSecondTryState`, `showSecondaryPaywallForPreview`, the delegate's preview branch, and the preview-specific diagnostic copy are all gone. The normal flow's `"\(trigger)_second_try"` lookup resolves the installed `helium_preview_trigger_second_try` entry naturally, and a miss surfaces the standard second-try-no-match diagnostic (which the preview gate always shows). The one preview-ism I kept is on the panel side where it belongs: the panel's onClose now checks the closing trigger, so a second try closing back onto the main preview doesn't release the panel's activity lock early.

One nuance to be aware of with the pure normal flow: the uuid lookup runs first, so if the previewed paywall's second try is also live under a real trigger, the preview can resolve and present that production entry (with its experiment info) instead of the synthetic preview entry. Same ambiguity production has when multiple triggers share a paywall — flagging in case you'd rather previews skip the uuid path.

Sources/Helium/HeliumCore/HeliumFetchedConfig.swift
JesusRojass: Extracted into previewEntryJSON in 43bc702 — both the main and second try JSON clones go through it now.

Sources/Helium/HeliumCore/Diagnostic/Mapper/DiagnosticContentMapper.swift
JesusRojass: Gone entirely in 43bc702 — the preview-specific diagnostic copy was deleted with the simplification, previews now surface the standard second-try-no-match record.

Sources/Helium/HeliumCore/HeliumActionsDelegate.swift
JesusRojass: It was possible (a second try paywall's own bundle can carry a second try CTA, which looped back into the preview branch) but it's moot now — 43bc702 removed that whole branch, so the case just follows the normal no-match path.

Sources/Helium/HeliumCore/HeliumFetchedConfig.swift
JesusRojass: Moved in 43bc702: previewSecondTryState is deleted outright per your suggestion, and isFallbackPreviewArmed now lives on HeliumControlPanelService with the config manager just writing it from the preview install paths.

Sources/Helium/HeliumCore/ControlPanel/HeliumControlPanelView.swift
```diff
@@ -337,7 +337,9 @@ struct HeliumControlPanelView: View {
         previewTask?.cancel()
         previewTask = Task {
             do {
+                async let secondTryTask = fetchSecondTryBundle(for: paywall)
                 let (bundleId, html) = try await HeliumControlPanelService.shared.fetchSingleBundle(bundleURL: bundleUrl)
+                let secondTryBundle = await secondTryTask
```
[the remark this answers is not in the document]
JesusRojass: The wait is already bounded: fetchSingleBundle sets request.timeoutInterval = 10, so the second try fetch can delay the preview by at most the same 10s ceiling the main bundle fetch has. A separate deadline race for an optional dev-tool fetch adds the kind of machinery this PR just removed, so leaving as is.

Sources/Helium/HeliumCore/ControlPanel/HeliumControlPanelView.swift
```diff
@@ -370,6 +373,35 @@ struct HeliumControlPanelView: View {
...
+                productIdsStripe: secondTry.stripeProductIds ?? [],
+                productIdsPaddle: secondTry.paddleProductIds ?? [],
+                shouldEnableScroll: secondTry.shouldEnableScroll
+            )
+        } catch {
+            HeliumLogger.log(.warn, category: .ui, "[HeliumControlPanel] Failed to load second try bundle for preview", metadata: [
+                "secondTryPaywall": secondTry.paywallName,
+                "error": "\(error)",
+            ])
+            return nil
```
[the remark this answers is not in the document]
JesusRojass: Intentional — the loadFailed/notConfigured state distinction was removed at reviewer request (see the review discussion) in favor of the plain second try flow: any unresolved second try surfaces the standard no-match diagnostic. A failed download is logged with the paywall name for anyone debugging, and losing the dedicated retry copy was the accepted trade-off for the simpler flow.

Sources/Helium/HeliumCore/HeliumFetchedConfig.swift
```diff
@@ -1166,29 +1250,39 @@ public class HeliumFetchedConfigManager {
...
+            configJSON["triggerToPaywalls"].dictionaryObject = triggers
+        }
         return configJSON
     }
+
+    private static func previewEntryJSON(clonedFrom sourceJSON: JSON, bundleUrl: String, shouldEnableScroll: Bool?) -> JSON {
+        var entryJSON = sourceJSON
+        entryJSON["resolvedConfig"]["baseStack"]["componentProps"]["bundleURL"] = JSON(bundleUrl)
+        entryJSON["resolvedConfig"]["baseStack"]["componentProps"]["shouldEnableScroll"] = JSON(shouldEnableScroll ?? true)
+        return entryJSON
```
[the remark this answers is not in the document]
JesusRojass: True but inert: the JSON mirror's only reader is getResolvedConfigJSONForTrigger, which extracts exactly resolvedConfig — additionalPaywallFields in the mirror is never read, and the typed config is the source of truth for paywallBundleUrl/webPaywallBundleUrl. The same asymmetry already exists for the main preview entry on main, so mirroring extra fields here would add maintenance surface for dead data. Leaving as is.

Sources/Helium/HeliumCore/HeliumActionsDelegate.swift
```diff
@@ -171,10 +171,6 @@ class HeliumActionsDelegate: ObservableObject {
     
     func showSecondaryPaywall(uuid: String?) {
         if !isLoading {
-            if trigger == HeliumFetchedConfigManager.HELIUM_PREVIEW_TRIGGER {
-                showSecondTryUnsupportedInPreviewDiagnostic()
-                return
-            }
             // Re-use same presentation context as underlying paywall. Integrator can check
             // isSecondTry to distinguish events.
             let presentationContext = paywallSession.presentationContext
```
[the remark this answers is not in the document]
JesusRojass: Good catch — this was the uuid nuance flagged in the top-level comment, and you're right that published second tries (the common case) would hit it. Fixed in d717a76: previews skip the uuid lookup with a single condition, so they resolve only the installed helium_preview_trigger_second_try entry via the normal trigger path. Production flow is untouched.

JesusRojass:
Closing the loop on the uuid nuance from the earlier comment: Bugbot flagged the same thing and rated it right — a published second try (the common case) would have resolved to its production entry from a preview. d717a76 settles it with a one-condition guard: previews skip the uuid lookup and resolve only the installed preview entry via the normal trigger path.

salami (approved):
Looks great thanks for the adjustments!

PR #329, opened 2026-08-13
Apply paywall preview server product details to fetched config and price map

JesusRojass:
## Problem

The server side of this now returns localized Stripe/Paddle product details on `/paywall-previews`, but the SDK dropped them on the floor: `HeliumControlPanelResponse` didn't declare the fields, so the decoder discarded them.

Even decoded, nothing consumed them. `fetchPaywalls` calls the `[String]` overload of `buildLocalizedPriceMap`, which only queries StoreKit — the server-price merge lives in the `config:` overload. And `setPreviewTriggerConfig` installs product ID lists into the cloned trigger without touching `stripeProducts` / `paddleProducts`.

Net effect: previewing a draft paywall renders no price, `HeliumPaymentProcessor.resolve(for:)` returns `.appStore` for its SKUs (routing purchases to StoreKit rather than the web processor), and Paddle checkout has no client token.

**Requires the corresponding bandit change to be deployed.** Without it these fields are simply absent and behavior is unchanged.

## Changes

- Declare `stripeProducts`, `paddleProducts`, `paddleClientToken` on `HeliumControlPanelResponse`. `ServerProductPrice` already matches the server's shape, so decoding needed no new types.
- New `HeliumFetchedConfigManager.mergePreviewServerProducts(...)`, writing to both places that matter for different consumers:
  - `fetchedConfig.stripeProducts` / `.paddleProducts` / `.paddleClientToken` — purchase routing, provider attribution, Paddle checkout prefetch
  - `localizedPriceMap` — what `DynamicWebView` actually renders, via `getLocalizedPriceMapForTrigger`
- Called from `fetchPaywalls` after the StoreKit price fetch.

## Notes for review

**Existing values win on collision.** On-launch resolved those against a real customer; the preview endpoint has no user and reports every intro offer as available. Letting it overwrite would show an ineligible tester a trial they can't actually get. It only ever fills gaps — which is the entire bug.

**Merged at fetch time rather than per-preview-arm,** since the maps are response-level and cover every version including the web handoff. This touches the shared config when the panel opens, which is safe: the added keys belong to no live trigger, and the control panel only opens in debug/sandbox/TestFlight builds.

## Testing

Full suite green — 380 tests, 0 failures. New `PreviewServerProductsTests` covers the config and price-map writes, the collision policy, filling a missing client token, the empty response, and a merge with no config present (no crash).

---

> [!NOTE]
> **Medium Risk**
> Touches product price maps and Paddle token used for purchase routing, but only overlays preview data and keeps live config values on collision. Limited to control-panel preview flows.
> 
> **Overview**
> Paywall previews now use Stripe/Paddle product details and a Paddle client token from `/paywall-previews`, so draft paywalls can show prices and route web checkout instead of dropping those fields.
> 
> Preview products live in a separate `previewServerProducts` store, not on `fetchedConfig`. Getters overlay them at read time so **live keys always win**, even if the preview arrives first. `applyServerProducts` skips canceled fetches so a late stale response cannot replace a newer one.
> 
> Paddle checkout prefetch reads the overlayed `paddleClientToken`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dc5e6dd9aba5a3b929675ebae21bf70715a52ad6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added support for Stripe and Paddle product pricing from preview configurations.
  * Added support for Paddle client tokens in preview configuration data.
  * Preview pricing and checkout settings now merge with fetched values while preserving live data precedence.
  * Paddle products and checkout settings now respect regional eligibility.

* **Bug Fixes**
  * Improved handling of empty, canceled, overlapping, or unavailable previews without disrupting existing settings.

* **Tests**
  * Added coverage for merging, localization, refresh behavior, precedence, eligibility, and missing configurations.

### DIFF
--- a/Sources/Helium/HeliumCore/HeliumFetchedConfig.swift
+++ b/Sources/Helium/HeliumCore/HeliumFetchedConfig.swift
@@ -211,6 +211,7 @@ public class HeliumFetchedConfigManager {
         shared.fetchedConfigJSON = nil
         shared.triggersWithSkippedBundleAndReason = []
         shared.localizedPriceMap = [:]
+        shared.previewInstalledProductKeys = []
         HeliumControlPanelService.shared.isFallbackPreviewArmed = false
     }
 
@@ -218,6 +219,7 @@ public class HeliumFetchedConfigManager {
     func injectConfigForTesting(_ config: HeliumFetchedConfig, json: JSON? = nil) {
         self.fetchedConfig = config
         self.fetchedConfigJSON = json
+        self.previewInstalledProductKeys = []
         self._downloadStatus = .downloadSuccess
     }
 
@@ -236,7 +238,8 @@ public class HeliumFetchedConfigManager {
     @HeliumAtomic private(set) var fetchedConfigJSON: JSON?
     @HeliumAtomic private(set) var triggersWithSkippedBundleAndReason: [(trigger: String, reason: PaywallUnavailableReason)] = []
     @HeliumAtomic private var localizedPriceMap: [String: LocalizedPrice] = [:]
-    
+    @HeliumAtomic private var previewInstalledProductKeys: Set<String> = []
+
     func fetchConfig(
         endpoint: String,
         apiKey: String,
@@ -345,6 +348,9 @@ public class HeliumFetchedConfigManager {
             self.fetchedConfig = newConfig
             self.fetchedConfigJSON = newConfigJSON
             triggersWithSkippedBundleAndReason = []
+            // A replaced config carries none of the preview's entries, so nothing of theirs
+            // is still owned — and its own values must not be treated as overwritable.
+            previewInstalledProductKeys = []
             
             if let stripeCustomerId = fetchedConfig?.stripeCustomerId {
                 applyCustomerIdFromConfig(stripeCustomerId, provider: .stripe)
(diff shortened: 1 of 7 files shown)


Sources/Helium/HeliumCore/HeliumFetchedConfig.swift
```diff
@@ -884,6 +890,73 @@ public class HeliumFetchedConfigManager {
         return allFound
     }
     
+    /// Existing values win: they were resolved against a real customer, while the preview endpoint
```
salami: What happens if a fallback opened before on-launch returns > paywall preview opened > then on-launch comes back. Unlikely, but do would the "true" on-launch values overwrite preview values in that case? Or do we risk preview values keeping precedence?

I wonder if we consider keeping the map separate and just tack on the map in `getStripeProductsPriceMap()` and `getPaddleProductsPriceMap()` to be safer?

JesusRojass: Oh good catch
buildLocalizedPriceMap(config:) merges on-launch prices existing-wins, so a preview value that landed first would've won permanently, not just till the next refresh.

addressed

salami (commented):
We don't need the eligiblePreviewPaddleProducts logic anymore. We decided bandit will send everything just with the forceExternalCheckoutSimulation flag

Also a note on a lot of these code comments -- I think they are more context than needed and exposing some of our internals in a public repo, plus they talk about cross-repo logic that if changed will go stale. The code comments should be simplified or dropped when not value-add.

Sources/Helium/HeliumCore/HeliumFetchedConfig.swift
JesusRojass: Done in 7e645c5 + dc5e6dd — the gate and isPaddleServedByOnLaunch are gone (reverted clean), previews go back to overlaying whatever the server sends, ready for the forceExternalCheckoutSimulation flag when bandit ships it. Also trimmed the comments across the PR down to the local contract, no more cross-repo narration.

salami (approved):
Looks good!

PR #332, opened 2026-08-14
Resolve server feature flags and gate JS-crash detection behind jsCrashFallback

JesusRojass:
## Why

The JS-crash detection that landed in #326 runs unconditionally on every render, and its recovery path can replace a showing paywall based on a heuristic — the review asked for a server-controlled kill-switch so a false-positive pattern in production can be disabled remotely instead of requiring an SDK release. Android already consumes the server's `featureFlags` map; iOS had no flag plumbing at all.

## What

- **`HeliumFeatureFlags`** — resolves the on-launch response's `featureFlags` map (served by cloudcaptainai/bandit-server-lite-phoenix#138), mirroring the Android SDK's semantics exactly: strict boolean coercion (string `"true"`, numbers, null and structures are logged and treated as off), undeclared keys ignored, and no client-side defaults — every flag is off unless the server sends `true`.
- **`jsCrashFallback` gate** — `DynamicWebView` only builds and injects the error hook when the flag is enabled, gating the whole detection path (hook → bridge → probes → recovery) at its root. WebContent-process-death handling stays ungated: that's a definitive OS signal, not a heuristic.
- Only `jsCrashFallback` is declared on iOS; Android's other flags map to concepts iOS already handles through existing config fields.

Matches the Android twin (cloudcaptainai/helium-android#336, which carries the same gate plus the no-defaults simplification).

## Testing

- 424 tests green, 9 new: flag-resolution semantics mirroring the Android suite (strict coercion, absent keys, empty/nil maps) and the on-launch decode contract for `featureFlags`.
- Rollout shape this enables: releases ship with detection dark, it's enabled server-side per org once validated, and can be killed remotely if `paywall_js_error_detected` outcomes show false positives.

HEL-6594

---

> [!NOTE]
> **Medium Risk**
> Changes paywall WebView failure and recovery paths behind a server flag; misconfiguration or rollout could leave users without automatic fallback while telemetry still fires.
> 
> **Overview**
> Adds **server-driven feature flags** from the on-launch `featureFlags` map, with strict boolean-only resolution (flags default off; invalid types are logged and treated as off).
> 
> **`jsCrashFallback`** is the first declared flag and acts as a remote kill-switch for paywall crash **recovery** behavior in `DynamicWebView`: the JS error-hook user script is only injected when enabled, and WebContent process termination no longer triggers `webViewLoadFail` reload/fallback unless the flag is on. Related **telemetry** (e.g. process-terminated tracking) remains emitted regardless of the flag.
> 
> Config plumbing includes `featureFlags` on `HeliumFetchedConfig` and `HeliumFetchedConfigManager.isFeatureEnabled(_:)`, plus unit tests for flag parsing and decode contract.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 82a3e8da1a95d5d80ba8b2351e324575f3e834d6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added server-controlled feature flags, including support for enabling or disabling JavaScript crash recovery.
  * Feature flags default to disabled and accept only valid Boolean values.
  * Added support for receiving feature flag settings from launch configuration.
* **Bug Fixes**
  * JavaScript error handling and web-content recovery now activate only when enabled.
* **Tests**
  * Added coverage for enabled, disabled, missing, invalid, and unknown feature flag values.

### DIFF
--- a/Sources/Helium/HeliumCore/Components/Web/DynamicWebView.swift
+++ b/Sources/Helium/HeliumCore/Components/Web/DynamicWebView.swift
@@ -279,8 +279,6 @@ struct DynamicWebView: View {
                 forMainFrameOnly: true
             )
 
-            // The server-driven flag gates the whole JS-crash detection path here at
-            // its root: with no hook installed, nothing downstream can ever fire.
             let errorHookScript: WKUserScript?
             if HeliumFetchedConfigManager.shared.isFeatureEnabled(.jsCrashFallback) {
                 errorHookScript = WKUserScript(
--- a/Sources/Helium/HeliumCore/Components/Web/DynamicWebView.swift
+++ b/Sources/Helium/HeliumCore/Components/Web/DynamicWebView.swift
@@ -200,6 +200,7 @@ struct DynamicWebView: View {
               ),
               scope: actionsDelegate.observabilityScope
           )
+          guard HeliumFetchedConfigManager.shared.isFeatureEnabled(.jsCrashFallback) else { return }
           webViewLoadFail(reason: "WebContentProcessTerminated", kind: .processTerminated)
       }
       .onReceive(NotificationCenter.default.publisher(for: .heliumWebCheckoutProcessingChanged)) { notification in
--- a/Sources/Helium/HeliumCore/Components/Web/DynamicWebView.swift
+++ b/Sources/Helium/HeliumCore/Components/Web/DynamicWebView.swift
@@ -279,11 +279,18 @@ struct DynamicWebView: View {
                 forMainFrameOnly: true
             )
 
-            let errorHookScript = WKUserScript(
-                source: WebViewRenderGuard.errorHookScriptSource(loadToken: loadToken),
-                injectionTime: .atDocumentStart,
-                forMainFrameOnly: true
-            )
+            // The server-driven flag gates the whole JS-crash detection path here at
+            // its root: with no hook installed, nothing downstream can ever fire.
+            let errorHookScript: WKUserScript?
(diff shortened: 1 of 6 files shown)


salami (approved):
**Approving** — the flag implementation looks solid: strict boolean coercion, default-off, per-load resolution, and good test coverage. Root-gating the injection is the right call; "flag off = zero new JS in any customer's page" is the strongest safety statement we can make.

One change I'd like before merge: **put the process-termination reaction under the flag too.** I know it's deliberately ungated (the enum doc comment says so), and the signal *is* definitive — but the reaction isn't risk-free. The termination observer isn't visibility-gated, and iOS kills WebContent processes for backgrounded apps under memory pressure; repeated terminations can walk the ladder to exhaustion and dismiss the paywall (plus fire an open-fail event) while the user isn't even in the app. As written, that's the only piece of the new behavior with no remote off-switch. Flag off should mean "the SDK behaves exactly like the previous release."

Concretely: in the `.onReceive(.webViewProcessTerminated)` handler, keep emitting `PaywallWebProcessTerminated` unconditionally (that's exactly the data that tells us whether background terminations are common enough to worry about), and only call `webViewLoadFail(kind: .processTerminated)` when `jsCrashFallback` is enabled. Update the enum doc comment to match — something like: the flag gates everything that can act on a paywall from the new detection paths (hook injection, termination reload), while telemetry and deterministic load-lifecycle fixes stay unconditional.

> 🤖 Written with [Claude Code](https://claude.com/claude-code) (paired with @kylegorlick).

PR #337, opened 2026-08-25
Pass preview config and environment into the external checkout

JesusRojass:
Stacked on the preview-settings branch — this diff is only the checkout wiring. GitHub will retarget once that PR merges.

- Parses top-level `forceExternalCheckoutSimulation` from `/preview-paywalls` (absent → true) to seed and lock the purchase-flow selection, and sends `userContext.locale.{currentCountry, storeCountryCode}` on the request so the server can evaluate the same signals on-launch uses.
- ctx now always carries `environment` (the app-receipt environment, verbatim: DEBUG / SANDBOX / PRODUCTION) and the `trigger` name; preview runs add `forceSimulatedFlow` from the selection. Field names match the bundler contract in cloudcaptainai/bundler-service#126.
- A Paddle preview with the CA toggle on appends `helium_force_california=true` to the checkout URL (the bundle gates on the param's presence); toggle off leaves the URL byte-identical.

The bundle acts on these once bundler-service#125/#126 deploy; unknown ctx fields pass through harmlessly today.

---

> [!NOTE]
> **Medium Risk**
> Changes checkout ctx/query for all external opens (not only previews) and touches California compliance paths in Paddle prefetch and preview URL building.
> 
> **Overview**
> Extends **control-panel preview paywalls** so the server can drive eligibility: `/preview-paywalls` responses decode **`forceExternalCheckoutSimulation`** (missing → simulated) and **`forcePaddleCaConsentModal`**, and the fetch POST now includes **`userContext.locale`** with device region and App Store country. Those flags seed **`HeliumPreviewConfigurationStore`**, including locking the California consent toggle when the server marks a CA device.
> 
> **External web checkout** always embeds **`environment`** (receipt environment) and **`trigger`** in compressed ctx. Preview runs also set **`forceSimulatedFlow`** from the tester’s purchase mode, and Paddle previews with CA consent enabled append **`helium_force_california=true`** on the URL. **Paddle prefetch** skips the California IP block when the run is a preview so testers can exercise consent and checkout.
> 
> Adds **`WebCheckoutPreviewConfigTests`** and updates response fixtures for the new model fields.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9398d1a6472cfad71529e95a2e86dac9f540616d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/Helium/HeliumCore/ControlPanel/HeliumPreviewConfigurationView.swift
+++ b/Sources/Helium/HeliumCore/ControlPanel/HeliumPreviewConfigurationView.swift
@@ -190,7 +190,7 @@ struct HeliumPreviewConfigurationView: View {
                 .toggleStyle(SwitchToggleStyle(tint: .blue))
                 .disabled(store.forcePaddleCaConsentModal)
                 if store.forcePaddleCaConsentModal {
-                    Label("Your device is in California, so the consent modal always shows.", systemImage: "lock.fill")
+                    Label("Paddle checkouts only. Your device is in California, so the consent modal always shows.", systemImage: "lock.fill")
                         .font(.caption)
                         .foregroundColor(.secondary)
                         .fixedSize(horizontal: false, vertical: true)
--- a/Sources/Helium/HeliumCore/ControlPanel/HeliumPreviewConfigurationView.swift
+++ b/Sources/Helium/HeliumCore/ControlPanel/HeliumPreviewConfigurationView.swift
@@ -188,6 +188,7 @@ struct HeliumPreviewConfigurationView: View {
                         .foregroundColor(.primary)
                 }
                 .toggleStyle(SwitchToggleStyle(tint: .blue))
+                .disabled(store.forcePaddleCaConsentModal)
                 Text("Paddle checkouts only. Real purchases from California always show the modal.")
                     .font(.caption)
                     .foregroundColor(.secondary)
--- a/Sources/Helium/HeliumCore/ControlPanel/HeliumPreviewConfigurationView.swift
+++ b/Sources/Helium/HeliumCore/ControlPanel/HeliumPreviewConfigurationView.swift
@@ -189,10 +189,17 @@ struct HeliumPreviewConfigurationView: View {
                 }
                 .toggleStyle(SwitchToggleStyle(tint: .blue))
                 .disabled(store.forcePaddleCaConsentModal)
-                Text("Paddle checkouts only. Real purchases from California always show the modal.")
-                    .font(.caption)
-                    .foregroundColor(.secondary)
-                    .fixedSize(horizontal: false, vertical: true)
+                if store.forcePaddleCaConsentModal {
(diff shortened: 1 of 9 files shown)


salami (commented):
Looking really solid just a couple asks
- Can we add a some kinda indicator if CA consent modal toggle is forced on? Like a "as a CA user you must see the modal" or something?
- In PaddleCheckoutPrefetchCoordinator we have a feature flagged zip code check that fails if CA detected. I think for the paywall previews experience we should also skip that check even if feature flag is off?
- If simulated option is chosen, I suppose we could still allow the CA toggle to be set OFF even if a CA user, but I don't know maybe that is overcomplicating things and better to just force it to show.

Sources/Helium/HeliumCore/ControlPanel/HeliumPreviewConfigurationView.swift
```diff
@@ -189,10 +189,17 @@ struct HeliumPreviewConfigurationView: View {
                 }
                 .toggleStyle(SwitchToggleStyle(tint: .blue))
                 .disabled(store.forcePaddleCaConsentModal)
-                Text("Paddle checkouts only. Real purchases from California always show the modal.")
-                    .font(.caption)
-                    .foregroundColor(.secondary)
-                    .fixedSize(horizontal: false, vertical: true)
+                if store.forcePaddleCaConsentModal {
+                    Label("Your device is in California, so the consent modal always shows.", systemImage: "lock.fill")
```
salami: I think we still need to clarify that this is Paddle maybe prepend "Paddle checkouts only. "?

salami (approved):
Added one more minor comment to consider. Looks good!

Sources/Helium/HeliumCore/ControlPanel/HeliumPreviewConfigurationView.swift
JesusRojass: Resolved!

Sources/Helium/HeliumCore/StripeCheckout/ExternalWebCheckoutManager.swift
```diff
@@ -315,6 +315,15 @@ public class ExternalWebCheckoutManager: NSObject {
 
         ctx["iosBundleId"] = Bundle.main.bundleIdentifier ?? "unknown"
 
+        ctx["environment"] = AppReceiptsHelper.shared.environment.rawValue.uppercased()
```
salami, line 318: Doing some testing here and realizing this value is actually passed under the analytics payload (`ctx["analytics"] = analyticsJSON`) above, so we do not need to pass this explicitly as I originally suggested.

salami: actually let's keep it as-is! Cleaner this way; bundler has easier access instead of having to dig through analytics payload

PR #343, opened 2026-09-02
Surface entitled products, source, and expiration in already-entitled skip diagnostics

JesusRojass:
## What

When a paywall is skipped because the user is already entitled, the diagnostic modal, log line, and copy-report now say *why*: the entitled product IDs, where the entitlement comes from (App Store / Paddle / Stripe), and the expiration date when available.

Example body suffix: `This user is entitled to: com.app.pro.yearly (App Store, expires Mar 3, 2026); prod_123 (Stripe).`

## How

- New internal `EntitledProductDetail` model (product ID, source, expiration date).
- `HeliumEntitlementsManager.entitledProductDetailsForPaywall(trigger:)` resolves the paywall's product IDs against cached StoreKit transactions (with `expirationDate`), persisted entitlements, and the labeled Paddle/Stripe/custom third-party sources. The existing source list was refactored to carry source labels.
- The skip funnel in `HeliumPaywallDelegateWrapper` fetches details for `.alreadyEntitled` before mapping, and adds an `entitledProducts` key to the structured log metadata.
- `DiagnosticContentMapper.mapSkip` accepts the details (defaulted) and appends them to the body, so the modal, log line, and support report all pick them up from the one content record.

## Notes

- Best-effort: when details cannot be resolved (paywalls not loaded, subscription-group match), the copy is unchanged from before.
- Targeting-holdout display untouched.
- `PaywallSkippedEvent` analytics payloads unchanged.

## Testing

- `xcodebuild build` (iOS Simulator destination) clean; existing unit tests unaffected (mapper signature defaulted).

---

> [!NOTE]
> **Low Risk**
> Changes are limited to diagnostic copy and logging for the already-entitled skip path; entitlement gating and analytics payloads are unchanged.
> 
> **Overview**
> When a paywall is skipped for **already entitled**, diagnostics now explain **which** products blocked the show—not just the generic message.
> 
> A new **`EntitledProductDetail`** model carries product key, purchase source (App Store / Paddle / Stripe / third-party), and optional expiration. **`HeliumEntitlementsManager`** aggregates entitled products from persisted StoreKit data, live transactions, and labeled third-party sources, then splits them into products on the trigger’s paywall vs. other active entitlements.
> 
> **`DiagnosticContentMapper.mapSkip`** accepts those lists (defaults keep other skip reasons unchanged) and appends human-readable lines to the modal/log/support copy. **`HeliumPaywallDelegate`** loads details asynchronously for `.alreadyEntitled` skips and adds an **`entitledProducts`** comma-separated key list to structured log metadata.
> 
> If paywalls aren’t loaded or product IDs can’t be resolved, behavior stays the same as before.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 97ec01e5c91a95528313edaf78615d793c05ab0b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **Diagnostics**
  - Paywall skip diagnostics now include entitlement details when access is already granted.
  - Details may include the product identifier, purchase source, and expiration date.

- **Bug Fixes**
  - Improved entitlement detection across App Store and supported third-party purchases.
  - Prevented duplicate product entries from multiple entitlement sources.
  - Improved handling of previously saved entitlements while transaction data loads.

### DIFF
--- a/Sources/Helium/HeliumCore/EntitledProductDetail.swift
+++ b/Sources/Helium/HeliumCore/EntitledProductDetail.swift
@@ -1,0 +1,19 @@
+//
+//  EntitledProductDetail.swift
+//  Helium
+//
+
+import Foundation
+
+enum EntitledProductSource: String {
+    case appStore = "App Store"
+    case paddle = "Paddle"
+    case stripe = "Stripe"
+    case thirdParty = "Third-party"
+}
+
+struct EntitledProductDetail: Equatable {
+    let productId: String
+    let source: EntitledProductSource
+    let expirationDate: Date?
+}
--- a/Sources/Helium/HeliumCore/EntitledProductDetail.swift
+++ b/Sources/Helium/HeliumCore/EntitledProductDetail.swift
@@ -13,7 +13,7 @@ enum EntitledProductSource: String {
 }
 
 struct EntitledProductDetail: Equatable {
-    let productId: String
+    let heliumProductKey: String
     let source: EntitledProductSource
     let expirationDate: Date?
 }
--- a/Sources/Helium/HeliumCore/HeliumEntitlementsManager.swift
+++ b/Sources/Helium/HeliumCore/HeliumEntitlementsManager.swift
@@ -718,12 +718,3 @@ actor HeliumEntitlementsManager {
     }
 }
 
-
-extension PaywallPresentationConfig {
-    func resolveIsEntitled(trigger: String) async -> Bool? {
-        if let isEntitledOverride {
-            return await isEntitledOverride()
-        }
-        return await Helium.entitlements.hasEntitlementForPaywall(trigger: trigger)
-    }
-}
--- a/Sources/Helium/HeliumCore/HeliumEntitlementsManager.swift
+++ b/Sources/Helium/HeliumCore/HeliumEntitlementsManager.swift
@@ -305,26 +305,26 @@ actor HeliumEntitlementsManager {
         guard Helium.shared.paywallsLoaded() else { return [] }
         let paywallInfo = HeliumFetchedConfigManager.shared.getPaywallInfoForTrigger(trigger)
             ?? HeliumFallbackViewManager.shared.getFallbackInfo(trigger: trigger)
-        let productIds = paywallInfo?.productIdsIncludingWebProductIds ?? []
-        guard !productIds.isEmpty else { return [] }
+        let productKeys = paywallInfo?.productIdsIncludingWebProductIds ?? []
+        guard !productKeys.isEmpty else { return [] }
 
-        var details: [EntitledProductDetail] = []
+        var entitledProducts: [EntitledProductDetail] = []
 
         if cache.lastTransactionsLoadedTime == nil {
             for persisted in cache.persistedEntitlements
-            where persisted.appearsValid() && productIds.contains(persisted.productID) {
-                details.append(EntitledProductDetail(
-                    productId: persisted.productID,
+            where persisted.appearsValid() && productKeys.contains(persisted.productID) {
+                entitledProducts.append(EntitledProductDetail(
+                    heliumProductKey: persisted.productID,
                     source: .appStore,
                     expirationDate: persisted.expirationDate
                 ))
             }
         }
-        if details.isEmpty {
+        if entitledProducts.isEmpty {
             let entitlements = await getCachedEntitlements()
-            for transaction in entitlements where productIds.contains(transaction.productID) {
-                details.append(EntitledProductDetail(
-                    productId: transaction.productID,
+            for transaction in entitlements where productKeys.contains(transaction.productID) {
+                entitledProducts.append(EntitledProductDetail(
+                    heliumProductKey: transaction.productID,
                     source: .appStore,
                     expirationDate: transaction.expirationDate
                 ))
@@ -333,18 +333,18 @@ actor HeliumEntitlementsManager {
 
         for (label, source) in labeledThirdPartySources {
             let entitledIds = await source.purchasedHeliumProductIds()
-            for productId in productIds
-            where entitledIds.contains(productId)
-                && !details.contains(where: { $0.productId == productId }) {
-                details.append(EntitledProductDetail(
-                    productId: productId,
+            for productKey in productKeys
+            where entitledIds.contains(productKey)
+                && !entitledProducts.contains(where: { $0.heliumProductKey == productKey }) {
+                entitledProducts.append(EntitledProductDetail(
+                    heliumProductKey: productKey,
                     source: label,
                     expirationDate: nil
                 ))
             }
         }
 
-        return details
+        return entitledProducts
     }
 
     func hasAnyActiveSubscription(includeNonRenewing: Bool) async -> Bool {
--- a/Sources/Helium/HeliumCore/HeliumEntitlementsManager.swift
+++ b/Sources/Helium/HeliumCore/HeliumEntitlementsManager.swift
@@ -14,18 +14,22 @@ actor HeliumEntitlementsManager {
     nonisolated let paddleEntitlementsSource = PaddleEntitlementsSource()
     nonisolated let stripeEntitlementsSource = StripeEntitlementsSource()
 
-    private var allThirdPartySources: [ThirdPartyEntitlementsSource] {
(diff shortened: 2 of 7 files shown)


Sources/Helium/HeliumCore/HeliumEntitlementsManager.swift
```diff
@@ -297,6 +301,52 @@ actor HeliumEntitlementsManager {
...
+            for persisted in cache.persistedEntitlements
+            where persisted.appearsValid() && productIds.contains(persisted.productID) {
+                details.append(EntitledProductDetail(
+                    productId: persisted.productID,
+                    source: .appStore,
+                    expirationDate: persisted.expirationDate
+                ))
+            }
+        }
+        if details.isEmpty {
```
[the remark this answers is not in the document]
JesusRojass: Skipping this one. The two sources never coexist: persisted entitlements are only the stand in before StoreKit transactions load, and they are cleared as soon as transactions arrive. The persisted first flow here mirrors the existing pattern in purchasedProductIds(), so the diagnostic reflects the same data view the skip decision itself used.

Sources/Helium/HeliumCore/HeliumEntitlementsManager.swift
```diff
@@ -668,3 +718,12 @@ actor HeliumEntitlementsManager {
...
 
+
+extension PaywallPresentationConfig {
+    func resolveIsEntitled(trigger: String) async -> Bool? {
+        if let isEntitledOverride {
+            return await isEntitledOverride()
+        }
+        return await Helium.entitlements.hasEntitlementForPaywall(trigger: trigger)
+    }
+}
```
salami: This leaked in from #344 right? Can you remove from this PR

Sources/Helium/HeliumCore/HeliumEntitlementsManager.swift
```diff
@@ -297,6 +301,52 @@ actor HeliumEntitlementsManager {
         return result
     }
     
+    func entitledProductDetailsForPaywall(trigger: String) async -> [EntitledProductDetail] {
+        guard Helium.shared.paywallsLoaded() else { return [] }
+        let paywallInfo = HeliumFetchedConfigManager.shared.getPaywallInfoForTrigger(trigger)
+            ?? HeliumFallbackViewManager.shared.getFallbackInfo(trigger: trigger)
```
salami: What if we thread paywallSession through logPaywallSkip then we can grab paywallInfo from there instead of re-extracting paywall info here? Less chance of drift

Sources/Helium/HeliumCore/HeliumEntitlementsManager.swift
```diff
@@ -297,6 +301,52 @@ actor HeliumEntitlementsManager {
...
+        let productIds = paywallInfo?.productIdsIncludingWebProductIds ?? []
+        guard !productIds.isEmpty else { return [] }
+
+        var details: [EntitledProductDetail] = []
```
salami: nit - `details` is kinda vague, could name it more specific

Sources/Helium/HeliumCore/EntitledProductDetail.swift
```diff
@@ -0,0 +1,19 @@
...
+
+enum EntitledProductSource: String {
+    case appStore = "App Store"
+    case paddle = "Paddle"
+    case stripe = "Stripe"
+    case thirdParty = "Third-party"
+}
+
+struct EntitledProductDetail: Equatable {
+    let productId: String
```
salami: can we name this `heliumProductKey`? To denote it as a canonical identifier for both StoreKit and Stripe/Paddle. I know the codebase doesn’t follow this pattern well yet, but I’d like to start differentiating our product key (which can be composite `product_id:price_id` for Stripe/Paddle) from a raw platform “product id”.

salami (approved):
Overall looks good, just added a few comments. And one overall thought: I wonder if we should list out _all_ products a user is entitled to. I feel like that might be useful information for developer/tester. We could potentially put this paywall's products first for emphasis but just listing them all out might be simpler and as effective. What do you think?

Also - does the diagnostic scroll in case this extra content starts growing too much vertically?

And on another note, both `entitledProductDetailsForPaywall` and `hasEntitlementForPaywall` re-extract `paywallInfo`. This is pre-existing, but I think would be cleaner to thread the exact paywall's products through. I created https://linear.app/tryhelium/issue/HEL-6833/hasentitlementforpaywall-should-not-re-derive-products

Sources/Helium/HeliumCore/HeliumEntitlementsManager.swift
JesusRojass: Yes, sorry addressed

salami:
@JesusRojass what do you think about my previous top-level comment on potentially listing full entitlements?

Sources/Helium/HeliumCore/HeliumPaywallDelegate.swift
```diff
@@ -361,12 +361,37 @@ class HeliumPaywallDelegateWrapper {
...
     ) {
-        let content = Self.diagnosticContentMapper.mapSkip(skipReason)
+        guard skipReason == .alreadyEntitled else {
+            emitSkipDiagnostic(trigger: trigger, skipReason: skipReason, entitledProducts: [], otherEntitlements: [], logMetadata: logMetadata)
+            return
+        }
+        Task {
+            let entitledProducts = await HeliumEntitlementsManager.shared.entitledProductDetailsForPaywall(trigger: trigger)
+            let otherEntitlements = await HeliumEntitlementsManager.shared.allEntitledProductDetails()
+                .filter { detail in !entitledProducts.contains(detail) }
```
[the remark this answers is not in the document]
JesusRojass: Skipping. That only happens in the pre load window where the paywall's products aren't resolved yet, and listing the full entitlements there is still the useful information for a tester. Suppressing it would hide data to fix a phrasing nit.

salami (approved):
🤖 Written with Claude Code

Approving — this is in good shape. The paywall-first + "Also entitled to" split reads well and the `heliumProductKey` rename is a nice cleanup.

One non-blocking suggestion on `logPaywallSkip`: `allEntitledProductDetails()` runs twice per skip — once inside `entitledProductDetailsForPaywall(trigger:)` and again directly for `otherEntitlements`. That doubles the third-party `purchasedHeliumProductIds()` calls (which can be network-backed), and because the two lists come from two separate snapshots, the `!entitledProducts.contains(detail)` exclusion (full-struct `Equatable`, including `expirationDate`) can momentarily drop a product into both sections or neither if state shifts between the calls.

Could fetch once and partition — e.g. a single manager method that returns both slices from one snapshot:

```swift
func entitledProductDetails(trigger: String) async -> (matching: [EntitledProductDetail], other: [EntitledProductDetail]) {
    let all = await allEntitledProductDetails()
    guard Helium.shared.paywallsLoaded(),
          let keys = paywallProductKeys(trigger), !keys.isEmpty else {
        return ([], all)
    }
    let matching = all.filter { keys.contains($0.heliumProductKey) }
    let other = all.filter { !matching.contains($0) }
    return (matching, other)
}
```

Halves the third-party calls and makes `other` a guaranteed complement of `matching`. Not a blocker — approving as-is.

Sources/Helium/HeliumCore/HeliumEntitlementsManager.swift
```diff
@@ -297,6 +301,57 @@ actor HeliumEntitlementsManager {
         return result
     }
     
+    func entitledProductDetails(trigger: String) async -> (matching: [EntitledProductDetail], other: [EntitledProductDetail]) {
+        let all = await allEntitledProductDetails()
+        guard Helium.shared.paywallsLoaded() else { return ([], all) }
+        let paywallInfo = HeliumFetchedConfigManager.shared.getPaywallInfoForTrigger(trigger)
+            ?? HeliumFallbackViewManager.shared.getFallbackInfo(trigger: trigger)
```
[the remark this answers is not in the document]
JesusRojass: Skipping. This matches how hasEntitlementForPaywall in the same file already resolves paywall info, and these symbols are public only for the wrapper SDK migration. A bespoke internal accessor for one call would be churn; that cleanup belongs to the broader public to internal migration and HEL-6833.

PR #350, opened 2026-09-09
HEL-6886 Add host-owned dismissal opt-out for embedded paywalls

salami:
## What

Adds an opt-out so an app hosting the embedded `HeliumPaywall` view can own dismissal instead of the SDK.

```swift
HeliumPaywall(trigger: "onboarding", eventHandlers: .withHandlers(
    onPurchaseSucceeded: { _ in /* advance yourself */ },
    onDismissed: { _ in /* pop yourself */ }
)) { reason in
    Text("no show: \(reason.description)")
}
.heliumDismissBehavior(.manual)
```

## Why

When the embedded paywall is placed in host-owned navigation (e.g. a real step in a `NavigationStack`), the SDK's post-purchase dismiss (`presentationMode.dismiss()`) pops that step out from under the host. There was no way to keep the paywall in place and drive navigation from the app. This is the natural counterpart to the embedded API — the host already owns presentation, so it can now own dismissal too (matching the Flutter embedded behavior).

## What changed

- New public `HeliumPaywallDismissBehavior` enum (`.automatic` default / `.manual`) and `View.heliumDismissBehavior(_:)` modifier, backed by an internal environment value — co-located with the embedded view in `HeliumPaywall.swift`.
- `DynamicBaseTemplateView` reads the behavior and, when `.manual`, suppresses both embedded dismiss channels: the `dismissAction` (purchase / restore / close button) and the `InlinePaywallDismissRegistry` registration (external Stripe/Paddle web checkout).

## Scope & behavior

- **Embedded only.** Gated on `viewType == .embedded`, so presented and `.heliumPaywall`-triggered paywalls keep their existing SDK-owned dismissal even if the modifier is set higher in the view tree.
- **Events still fire.** Only the dismiss action is suppressed — `onPurchaseSucceeded`, restore, and `onDismissed`/`PaywallDismissedEvent` all still emit, so the host has the signals it needs.
- **Close button caveat.** With `.manual`, the paywall's close button becomes a no-op unless the host dismisses the view from `onDismissed`. This is intended and documented on the modifier.

## Testing

Manually verified against an embedded paywall pushed as a `NavigationStack` step (example intentionally not included in this PR): on purchase the paywall stays put and a host "thank you" renders over it; close routes through `onDismissed`. Default (`.automatic`) behavior is unchanged.

## Follow-up

[...]

PR #351, opened 2026-09-09
Add onEntitled to embedded HeliumPaywall and heliumPaywall modifier

JesusRojass:
## What

Adds an `onEntitled` handler to the embedded `HeliumPaywall` view and the `.heliumPaywall` modifier, matching `presentPaywall` and the Android embedded API.

```swift
HeliumPaywall(
    trigger: "onboarding",
    config: PaywallPresentationConfig(dontShowIfAlreadyEntitled: true),
    onEntitled: { event in /* advance yourself */ }
) { reason in
    Text("no show: \(reason.description)")
}
```

## Why

Hosts of the embedded view had to reconstruct "user is now entitled" from `onPurchaseSucceeded` / `onPurchaseRestored` and handle the already-entitled skip separately through `whenPaywallNotShown(.alreadyEntitled)`. With host-owned dismissal (#350) the host drives navigation after purchase, and `onEntitled` is the signal for that. The SDK already marked embedded sessions entitled on purchase, but only the presented close path consumed the mark, so those entries leaked.

## What changed

- `HeliumPaywall` inits and both `heliumPaywall(...)` overloads take `onEntitled: ((PaywallEntitledEvent) -> Void)? = nil`, passed through `PaywallPresentationContext`.
- `HeliumPaywallPresenter.dispatchEntitledEvent(forSessionId:presentationContext:)` consumes the marked event and calls the handler on the main actor. The presented close path uses it, and the inline SDK dismiss closure in `DynamicBaseTemplateView` calls it for embedded and triggered paywalls before the `.manual` guard, so it fires at the point the SDK would dismiss whether or not the view is removed. That closure is the single path for webview purchase and restore, already-entitled purchase, and external web checkout completion.
- With `dontShowIfAlreadyEntitled` and an entitled user, `whenPaywallNotShown(.alreadyEntitled)` renders as before and a provided `onEntitled` also receives `.skipped` with the same `PaywallSkippedEvent` sent to analytics, through the same `handleAlreadyEntitledSkip` path the presented flow uses.
- `fireEvent` only marks a session entitled when the context has an `onEntitled` handler. That removes the leak for handler-less sessions and for the detached pending-purchase session on every surface.
- `HeliumPaywallPresentationState.isInline` replaces the repeated `viewType != .presented` checks.

## Scope & behavior

- Presented paywalls: unchanged, `onEntitled` on close.
- Embedded and triggered: `onEntitled` is scheduled on the main actor at completion, before the view's `PaywallCloseEvent`.
- Pending (Ask to Buy) purchases that resolve after the paywall is gone still do not deliver `onEntitled`, as before, and no longer leave a mark behind.

## Testing

- Unit suite on the iPhone 17 simulator, 485 tests passing. New coverage: dispatch fires once and consumes, consumes without a handler, no call without a mark, no mark without a handler.
- Example app's embedded paywall wires `onEntitled`, ready for device testing of purchase, restore, and the already-entitled skip under `.automatic` and `.heliumDismissBehavior(.manual)`.

---

> [!NOTE]
> **Medium Risk**
> Touches purchase/entitlement callback timing across embedded, triggered, and presented paywalls; behavior is well tested but hosts may depend on the new callback firing at dismiss time.
> 
> **Overview**
> Adds an optional **`onEntitled`** callback to embedded **`HeliumPaywall`** and the **`.heliumPaywall`** modifier so hosts get one signal after purchase, restore, already-entitled purchase, or an already-entitled skip when **`dontShowIfAlreadyEntitled`** is on—aligned with **`presentPaywall`** and host-owned dismissal.
> 
> Entitlement delivery is centralized in **`HeliumPaywallPresenter.dispatchEntitledEvent`**, used on presented close and in the inline dismiss path (**`HeliumBaseTemplateView`**) before the manual-dismiss guard, so embedded/triggered paywalls fire **`onEntitled`** even when the host keeps the view visible. **`fireEvent`** only marks sessions entitled when a handler is registered, fixing leaked marks for handler-less and pending-purchase sessions. The embedded **`.alreadyEntitled`** path now routes through **`handleAlreadyEntitledSkip`** instead of firing analytics only.
> 
> Tests cover dispatch-once/consume behavior and no-mark-without-handler; the example app logs **`onEntitled`** for the embedded paywall.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 697ce55dc53e2385516dade0e0e2421e222fea85. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added optional entitlement callbacks to paywall views and modifiers.
  * Entitlement events now trigger the callback for inline and presented paywalls.
  * Added entitlement handling for already-entitled users.

* **Bug Fixes**
  * Sessions without an entitlement callback are no longer incorrectly marked as entitled.
  * Inline paywall registration and dismissal behavior is now handled correctly.

### DIFF
--- a/HeliumExample/HeliumExample/ContentView.swift
+++ b/HeliumExample/HeliumExample/ContentView.swift
@@ -73,7 +73,12 @@ struct ContentView: View {
                 .accessibilityIdentifier("showPaywallEmbedded")
                 .fullScreenCover(isPresented: $showEmbeddedPaywall) {
                     HeliumPaywall(
-                        trigger: trigger, config: PaywallPresentationConfig(dontShowIfAlreadyEntitled: dontShowIfAlreadyEntitled)
+                        trigger: trigger,
+                        config: PaywallPresentationConfig(dontShowIfAlreadyEntitled: dontShowIfAlreadyEntitled),
+                        onEntitled: { event in
+                            print("[Helium Example] embedded onEntitled: \(event)")
+                            showEmbeddedPaywall = false
+                        }
                     ) { reason in
                         Text("no show embedded! \(reason.description)")
                     }
--- a/HeliumExample/HeliumExample/ContentView.swift
+++ b/HeliumExample/HeliumExample/ContentView.swift
@@ -77,7 +77,6 @@ struct ContentView: View {
                         config: PaywallPresentationConfig(dontShowIfAlreadyEntitled: dontShowIfAlreadyEntitled),
                         onEntitled: { event in
                             print("[Helium Example] embedded onEntitled: \(event)")
-                            showEmbeddedPaywall = false
                         }
                     ) { reason in
                         Text("no show embedded! \(reason.description)")
--- a/Sources/Helium/HeliumCore/HeliumBaseTemplateView.swift
+++ b/Sources/Helium/HeliumCore/HeliumBaseTemplateView.swift
@@ -51,12 +51,19 @@ public struct DynamicBaseTemplateView: View {
             // the view itself. Presented and triggered paywalls keep SDK-owned dismissal.
             let manualDismissal = dismissBehavior == .manual && presentationState.viewType == .embedded
             let dismissUnlessManual: () -> Void = {
+                if presentationState.isInline {
+                    let session = actionsDelegate.paywallSession
+                    HeliumPaywallPresenter.shared.dispatchEntitledEvent(
+                        forSessionId: session.sessionId,
+                        presentationContext: session.presentationContext
+                    )
+                }
                 guard !manualDismissal else { return }
                 dismiss()
             }
 
             actionsDelegate.setDismissAction(dismissUnlessManual)
-            if presentationState.viewType != .presented {
+            if presentationState.isInline {
                 InlinePaywallDismissRegistry.register(sessionId: actionsDelegate.paywallSession.sessionId, dismissUnlessManual)
                 if !presentationState.isOpen {
                     presentationState.isOpen = true
@@ -71,7 +78,7 @@ public struct DynamicBaseTemplateView: View {
             }
         }
         .onDisappear {
-            if presentationState.viewType != .presented {
+            if presentationState.isInline {
                 InlinePaywallDismissRegistry.unregister(sessionId: actionsDelegate.paywallSession.sessionId)
                 if presentationState.isOpen {
                     presentationState.isOpen = false
--- a/Sources/Helium/HeliumCore/HeliumPaywall.swift
+++ b/Sources/Helium/HeliumCore/HeliumPaywall.swift
@@ -43,8 +43,8 @@ public struct HeliumPaywall<PaywallNotShownView: View>: View {
     ///   - trigger: The trigger name to display a paywall for
     ///   - config: Additional configuration options
     ///   - eventHandlers: Optional event handlers for paywall lifecycle events
-    ///   - onEntitled: Called with the entitling event on purchase, restore, or an already-entitled purchase, regardless of `.heliumDismissBehavior`.
-    ///    With `config.dontShowIfAlreadyEntitled`, also called with `.skipped` when the paywall is skipped for an entitled user.
+    ///   - onEntitled: Called with the entitling event on purchase, restore, or an already-entitled purchase.
+    ///    Also called with `.skipped` when the paywall is skipped because `config.dontShowIfAlreadyEntitled` is true and the user is already entitled.
     ///   - whenPaywallNotShown: View to show when paywall is unavailable or skipped due to targeting/already-entitled
     public init(
         trigger: String,
@@ -74,8 +74,8 @@ public struct HeliumPaywall<PaywallNotShownView: View>: View {
     ///   - trigger: The trigger name to display a paywall for
     ///   - config: Additional configuration options
     ///   - eventHandlers: Optional event handlers for paywall lifecycle events
-    ///   - onEntitled: Called with the entitling event on purchase, restore, or an already-entitled purchase, regardless of `.heliumDismissBehavior`.
-    ///    With `config.dontShowIfAlreadyEntitled`, also called with `.skipped` when the paywall is skipped for an entitled user.
+    ///   - onEntitled: Called with the entitling event on purchase, restore, or an already-entitled purchase.
+    ///    Also called with `.skipped` when the paywall is skipped because `config.dontShowIfAlreadyEntitled` is true and the user is already entitled.
     ///   - loadingView: Custom view to show while paywall is loading
     ///   - whenPaywallNotShown: View to show when paywall is unavailable or skipped due to targeting/already-entitled
     public init<LoadingView: View>(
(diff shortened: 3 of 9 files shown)


HeliumExample/HeliumExample/ContentView.swift
```diff
@@ -73,7 +73,12 @@ struct ContentView: View {
                 .accessibilityIdentifier("showPaywallEmbedded")
                 .fullScreenCover(isPresented: $showEmbeddedPaywall) {
                     HeliumPaywall(
-                        trigger: trigger, config: PaywallPresentationConfig(dontShowIfAlreadyEntitled: dontShowIfAlreadyEntitled)
+                        trigger: trigger,
+                        config: PaywallPresentationConfig(dontShowIfAlreadyEntitled: dontShowIfAlreadyEntitled),
+                        onEntitled: { event in
+                            print("[Helium Example] embedded onEntitled: \(event)")
+                            showEmbeddedPaywall = false
```
salami: This shouldn't be needed unless manual dismissal is used. Let's remove this line in case agents look at this example code.

Sources/Helium/HeliumCore/HeliumPaywall.swift
```diff
@@ -105,12 +111,20 @@ public struct HeliumPaywall<PaywallNotShownView: View>: View {
                     .environment(\.heliumLoadTimeTakenMS, resolvedLoadTimeTakenMS)
                     .environment(\.heliumDynamicPaywallTraits, config.customPaywallTraits)
             case .noShow(let reason):
-                paywallNotShownView(reason)
-                    .onAppear {
-                        onPaywallUnavailable(reason: reason)
-                    }
+                if reason == .alreadyEntitled && handlesAlreadyEntitledSkip {
+                    EmptyView()
```
salami: I'm skeptical of adding an empty view. We've had this in the past and it is confusing. I'd lean towards leaving this as is even though it diverges a bit from the presented flow. paywallNotShownView can just show like it would have before these changes

Sources/Helium/HeliumCore/HeliumBaseTemplateView.swift
```diff
@@ -51,12 +51,19 @@ public struct DynamicBaseTemplateView: View {
             // the view itself. Presented and triggered paywalls keep SDK-owned dismissal.
             let manualDismissal = dismissBehavior == .manual && presentationState.viewType == .embedded
             let dismissUnlessManual: () -> Void = {
+                if presentationState.isInline {
+                    let session = actionsDelegate.paywallSession
+                    HeliumPaywallPresenter.shared.dispatchEntitledEvent(
+                        forSessionId: session.sessionId,
+                        presentationContext: session.presentationContext
+                    )
+                }
```
salami: I think, to keep this simpler, we can just call onEntitled in Helium paywall presenter. Like instead of markSessionAsEntitled and deferring the callback we can just do it right away for embedded paywalls.

The only reason we added that song and dance was because, for presented paywalls, we want consumers to not have to worry about whether the presenting view controller (e.g. presentFromViewController) is still displaying. For embedded paywall, that constraint does not exist. I think we can simplify things.

Sources/Helium/HeliumPaywallPresentation/HeliumPaywallPresenter.swift
```diff
@@ -483,14 +490,7 @@ class HeliumPaywallPresenter {
...
         dispatchOpenOrCloseEvent(openEvent: false, paywallVC: paywallVC)
-
-        // Call onEntitled if this session had a successful purchase/restore
-        let sessionId = paywallVC.paywallSession.sessionId
-        if let entitledEvent = consumeEntitledEvent(forSessionId: sessionId) {
-            Task { @MainActor in
-                paywallVC.presentationContext.onEntitled?(entitledEvent)
-            }
-        }
+        dispatchEntitledEvent(forSessionId: paywallVC.paywallSession.sessionId, presentationContext: paywallVC.presentationContext)
```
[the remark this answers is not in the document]
JesusRojass: This ordering is pre-existing on the presented path (main has the same two steps in dispatchCloseEvent) and matches the presentPaywall doc, which says onEntitled is called when the paywall closes. Flipping it would change behavior for every existing presentPaywall integrator, so leaving it as is here.

salami (commented):
🤖 Reviewed with Claude Code

Solid implementation — it handles the parts that are easy to get wrong:

- **`onEntitled` fires under `.manual` too**, since `dispatchEntitledEvent` runs before the dismiss guard — not just on auto-dismiss.
- **Both restore paths** (subscribe `.restored` and `restore-purchases`) flow through the same dismiss → dispatch, so neither is missed.
- **The already-entitled skip fires exactly once** — the `EmptyView`+`onChange` and `paywallNotShownView`+`onAppear` branches are mutually exclusive.
- Reusing `dispatchEntitledEvent` across the presented and inline paths keeps `viewType` out of the model layer (the inline consume is triggered from the view, where `isInline` legitimately lives). I looked at whether the mark→consume could be dropped for embedded and concluded it shouldn't — it's the right carrier for crossing the delegate → view boundary without leaking presentation into the session.

Two changes (details in inline comments):

1. **`showEmbeddedPaywall = false` in the example is redundant** under the default `.automatic` dismissal — the SDK already closes the cover, so it's only needed for `.manual`. I'd drop it so the example doesn't imply `onEntitled` handlers must dismiss.

2. **The `EmptyView()` for the already-entitled skip** — should-fix, for a concrete reason beyond the blank-view confusion: it forces the skip to fire from `.onChange(of: state)`, and `onChange` doesn't fire for an *initial* value. Today you're only safe because `dontShowIfAlreadyEntitled` resolves in two phases (`.checkingEntitlement` → `.alreadyEntitled`); if that ever resolved synchronously, `onEntitled(.skipped)` would silently never fire and you'd render blank. Rendering `paywallNotShownView` as before (its `onAppear` reliably fires the skip) is simpler and lets you delete both the `EmptyView` branch and the `.onChange` block. Minor tradeoff: the host's not-shown view may flash briefly before they navigate away in `onEntitled` — but that's their own view, not blank.

Nothing blocking except #2, which I'd treat as should-fix.

Sources/Helium/HeliumCore/HeliumPaywall.swift
```diff
@@ -71,12 +74,15 @@ public struct HeliumPaywall<PaywallNotShownView: View>: View {
     ///   - trigger: The trigger name to display a paywall for
     ///   - config: Additional configuration options
     ///   - eventHandlers: Optional event handlers for paywall lifecycle events
+    ///   - onEntitled: Called with the entitling event on purchase, restore, or an already-entitled purchase, regardless of `.heliumDismissBehavior`.
+    ///    With `config.dontShowIfAlreadyEntitled`, also called with `.skipped` when the paywall is skipped for an entitled user.
```
salami: This line is confusing to me. Maybe something like

> Also called when the paywall is skipped if config.dontShowIfAlreadyEntitled is true and the user is already entitled.

or just remove it?

Also thinking we should remove "regardless of `.heliumDismissBehavior`"  as it's arguably obvious enough that it should not be stated. I don't think an app integrator would expect heliumDismissBehavior to affect whether "onEntitled" is called.

salami (approved):
Looks good! Added one more note on a public-facing docstring to consider.

On a related note (not to be addressed here), part of me wonders if we should just rip out dontShowIfAlreadyEntitled support from the embedded paywall.

PR #353, opened 2026-09-15
HEL-6939 Add server-controlled in-app web checkout presentation

salami:
Paywalls can now specify how the External Web Checkout URL is presented. The server sends `webCheckoutBrowserStyle` per paywall in the on-launch response:

- `externalBrowser` — current behavior, hands off to the user's default browser
- `safariSheet` — `SFSafariViewController` as a page sheet over the paywall
- `safariFullScreen` — `SFSafariViewController` full screen

There is no local override, so presentation can be changed or reverted server-side without an app update. An absent or unrecognized value resolves to `externalBrowser`, and an unknown string decodes as unset rather than throwing, so one bad value cannot fail the surrounding paywall decode. Nothing changes until the server starts sending the field.

## Purchase completion

The external flow learns checkout is over when the app returns to the foreground. An in-app browser never backgrounds the app, so that signal never arrives.

- A success or cancel redirect closes the browser. Success closes it before the entitlement check, so the processing overlay — which renders on the paywall — is visible while the check runs.
- The browser reporting its own dismissal replaces the foreground return as the safety net. It is keyed to `viewDidDisappear` rather than the Safari dismiss button, since a swiped-down sheet never reaches `safariViewControllerDidFinish`.
- The foreground observer is armed for every style. The check it triggers is skipped only while our own browser is still on screen, which is the one moment a foreground return genuinely says nothing about checkout — so backgrounding mid-checkout still costs no entitlement refresh. The guard runs before the observer is torn down, so a wrong read costs one skipped check rather than every future one.

Dismissals the SDK initiates do not report back, so closing the browser on a redirect costs no extra entitlement call.

## Observability

`browserStyle` is now on all four web checkout events — browser open attempted, redirect received, purchase detected, purchase check exhausted — each resolved from its own observation's paywall session, so every arm is attributable end to end.

## Notes

- `SFSafariViewController` is hosted as a child of a container so its dismissal can be observed, and so the full screen style slides up rather than sideways; presented directly it installs its own transitioning delegate that overrides `modalTransitionStyle`.
- Only `externalBrowser` hands the user off to their default browser as the external purchase link entitlement expects. The Safari styles are experimental, intended for a single opt-in customer on the US storefront, and server-controlled specifically so they can be turned off without an app update.
- A purchase that completes server-side while the app is backgrounded, and whose redirect never fires, is detected when the user closes the browser — which they have to do to get back to the app.

Tested on device against Paddle and Stripe checkouts: dismiss button, swipe-to-dismiss, cancel redirect, and a completed purchase including the Apple Pay sheet.

---

> [!NOTE]
> **Medium Risk**
> Changes purchase-completion detection and session lifecycle for web checkout; mistakes could miss purchases or mis-attribute entitlements, though external-browser behavior remains the default fallback.
> 
> **Overview**
> Paywalls can now drive **how external web checkout opens** via server field `webCheckoutBrowserStyle`: default **external browser** (unchanged), **in-app Safari sheet**, or **full-screen Safari**. Unknown or missing values decode safely and still default to external.
> 
> Checkout no longer always uses `UIApplication.shared.open`. **`WebCheckoutPresenter`** routes URLs by style and hosts **`SFSafariViewController`** in a container (loading cover, dismiss observation). **`ExternalWebCheckoutManager`** resolves style per paywall session, tracks when an in-app browser is showing, and **skips foreground entitlement checks** while that UI is up. **Browser dismissal** and **redirect handling** now close in-app Safari, run purchase checks at the right time, and **drop observations** for abandoned in-app sessions on cancel/failure so stray entitlements are not attributed.
> 
> **Observability:** all four web checkout events include **`browserStyle`** for attribution.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dded920c5d56eea1a8fd3a8cca2350773fefb28e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Web checkout can open in an in-app Safari sheet or full-screen browser, in addition to the external browser.
  * Checkout presentation style can be configured remotely, with safe fallback behavior for unsupported values.
  * Paywall checkout information now includes the selected presentation style.

* **Observability**
  * Web checkout events now record the presentation style and purchase restoration status.
  * Checkout activity is tracked consistently across supported presentation styles.


JesusRojass (approved):
LGTM

Sources/Helium/HeliumCore/Models.swift
```diff
@@ -46,6 +46,7 @@ public struct HeliumPaywallInfo: Codable {
     var experimentInfo: JSON?  // New top-level field from server
     var additionalPaywallFields: JSON?
     var presentationStyle: HeliumPresentationStyle?
+    var webCheckoutBrowserStyle: WebCheckoutBrowserStyle?
```
Arhan13, line 49: Follow-up, not a bug in the on-launch path: because this is a typed top-level field, it rides along when `previewEntry` (`HeliumFetchedConfig.swift`) clones the source trigger's paywall info for a control panel preview. That function overrides the bundle URL, `webPaywallBundleUrl`, the five product lists, scroll and `forceShowFallback`, but not `webCheckoutBrowserStyle`, so a previewed paywall opens checkout with the style of whichever paywall is live on the source trigger. Its doc comment says nothing of the source paywall should leak into the preview.

The SDK cannot fix this alone today: phoenix's `paywall-previews` response does not include the style (noted on phoenix #157). Once it does, `previewEntry` should set it from the response and nil it when absent, the same way it already handles `webPaywallBundleUrl`. Until then, nil-ing it in `previewEntry` would at least make previews consistently use the default external browser. Production triggers are unaffected.

Arhan13 (commented):
Post-merge note from a cross-repo pass over HEL-6939 / HEL-6945 (this PR, phoenix #157, CML #1521). The on-launch decode and the single read site in ExternalWebCheckoutManager look correct. One follow-up inline, for the control panel preview path.

PR #357, opened 2026-09-16
[HEL-5834] Report browser Apple Pay readiness on launch for server targeting

Arhan13:
## Summary

Measures whether the browser can actually pay with Apple Pay, and reports it on launch so Bandit server targeting decides web versus in-app. The SDK does not route on it.

The measurement is one hidden 1x1 `WKWebView` loaded at the checkout origin, running the raw call the checkout page itself relies on:

```js
ApplePaySession.canMakePaymentsWithActiveCard('merchant.' + location.hostname + '.stripe')
```

On device this matched real Safari and live checkout in all three tested states (card present, no card, Safari "Check for Apple Pay" off), and answers in about 0.5 s. No Stripe.js, no publishable key.

true -> `.ready`, false -> `.notReady`, everything else -> `.unknown(Reason)` with `notMeasured`, `deviceCannotPay`, `noApplePayAPI`, `apiError`, `probeFailed`, `timedOut`, serialized as `unknown:timedOut` and carried in `webApplePayReadiness` on the launch payload.

Shape, per the plan's "Arhan thoughts" section:

```
opt-in -> persisted readiness loaded synchronously -> cached value sent on launch
       -> nonblocking refresh when web checkout is enabled -> result persisted for the next launch
```

- `Helium.config.enableWebApplePayReadiness` defaults to false. Off means never probe, `readiness()` returns `.ready`, launch reports `"ready"`, behavior unchanged.
- Persistence follows the store-country pattern through `HeliumStorage`, keyed with the origin it was measured at, since the merchant id derives from the checkout hostname. An unknown outcome is never persisted as a measurement, so the last measured answer survives a failed refresh.
- No local routing and no loading-budget wait. Readiness never changes which paywall is shown on device, which is why `webApplePayNotReady` and its diagnostic are gone.
- `WebApplePayProbe` is one-shot guarded across the JS result, navigation failure, web content process termination, and a 2 s timeout.
- `WebApplePayProbeCompleted` carries readiness reason, duration, timeout, failure reason, cache source, and whether the cached value matched what the probe then measured.

Native `hasApplePayCreditCard` / `hasApplePayDebitCard` are deliberately not part of this; they under-reported in production and excluded users who did pay.

Until the server-side targeting ruleset exists this ships as pure telemetry, so nobody is routed away from a dead Apple Pay checkout yet. Registering `webApplePayReadiness` as a targetable system variable is a separate CML change.

Stripe coverage is stacked on top in #358. Full Catalyst suite green.

Link to Devin session: https://app.devin.ai/sessions/3b80f13d2ed34019b898eaca123258e5
Open in Devin Desktop: https://app.devin.ai/desktop/session/3b80f13d2ed34019b898eaca123258e5?variant=devin
Requested by: @Arhan13

---

> [!NOTE]
> **Medium Risk**
> Opt-in feature touches launch/config timing (up to 2s first launch) and runs WKWebView Apple Pay probes, but defaults off and does not alter on-device paywall routing until server rules use the new field.
> 
> **Overview**
> Adds **opt-in browser Apple Pay readiness** (`Helium.config.enableWebApplePayReadiness`, default off) so launch targeting can distinguish “device supports Apple Pay” from “Wallet has an active card for the web checkout merchant.” When enabled and web checkout is configured, a hidden `WKWebView` probes `ApplePaySession.canMakePaymentsWithActiveCard` at the Paddle checkout origin; results are cached in `HeliumStorage`, reported as `webApplePayReadiness` on the user context payload, and refreshed once per launch. **First launch** may delay the paywall config fetch by up to **2s** (in parallel with store-country prefetch) until a measurement exists; opted-out apps still report `"ready"` and never probe.
> 
> Probe outcomes emit **`web_apple_pay_probe_completed`** telemetry (duration, timeout, cache correctness, launch wait). **Paywall routing is unchanged**—readiness is telemetry/targeting only. Includes test hooks on `ApplePayHelper` and broad unit coverage for probe, cache, and launch-wait behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit af952bab9c5d6c9669cb348ae231cf8746b8f1a6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added optional browser-based Apple Pay readiness detection during app initialization.
  - Checks run only when enabled and supported, with results cached and reused across launches.
  - Added configuration to enable or disable readiness checks; disabled by default.
  - Checkout context now reports web Apple Pay readiness alongside device Apple Pay availability.
  - Configuration loading may briefly wait for the first readiness measurement; later measurements run in the background.
  - Added diagnostics for probe outcomes, including readiness, duration, timeouts, cache usage, and failure details.


Sources/Helium/HeliumCore/StripeCheckout/WebApplePayAvailability.swift
```diff
@@ -0,0 +1,203 @@
...
+    /// A device that cannot do Apple Pay at all is left as `unknown` rather than
+    /// reported as not ready, so the value only ever reflects a real measurement.
+    func shouldProbe() -> Bool {
+        guard Helium.config.enableWebApplePayReadiness else { return false }
+        guard !Helium.config.webCheckoutProcessors.isEmpty else { return false }
+        guard ApplePayHelper.shared.canMakePayments() else { return false }
+        return !probeAttempted
+    }
+
+    func reset() {
```
salami: I don't think we should reset this. It provides no benefit and just means we have to add extra delays.

Sources/Helium/HeliumCore/StripeCheckout/WebApplePayAvailability.swift
```diff
@@ -0,0 +1,203 @@
...
+        generation += 1
+        storage.remove(forKey: Self.persistedReadinessKey)
+        storage.remove(forKey: Self.persistedOriginKey)
+    }
+
+    /// The browser evaluates Apple Pay against the origin serving checkout, and the merchant
+    /// identifier is derived from that hostname, so the probe has to load the same origin the
+    /// web paywall bundle is served from. Paddle and Stripe checkout share it.
+    func probeOrigin() -> URL? {
+        if let bundleOrigin = webPaywallBundleOrigin() {
```
salami: I think we can just always check against the production origin instead of doing it with an `if-else` here from webPaywallBundleOrigin? It simplifies things.

Sources/Helium/HeliumCore/StripeCheckout/WebApplePayAvailability.swift
```diff
@@ -0,0 +1,203 @@
...
+    func measurementMatchesCurrentOrigin() -> Bool {
+        guard let probedOrigin, let currentOrigin = probeOrigin() else { return true }
+        return probedOrigin == currentOrigin
+    }
+
+    /// Starts the one refresh this launch gets. Returns immediately; the result lands in the
+    /// cache and in storage when the probe completes.
+    func refreshIfNeeded() {
+        guard shouldProbe() else { return }
+        guard let origin = probeOrigin() else { return }
```
salami: I don't know if we necessarily need to do the observability event for an early return like this, but we should at least add a log.

Sources/Helium/HeliumCore/StripeCheckout/WebApplePayAvailability.swift
```diff
@@ -0,0 +1,203 @@
...
+    func apply(_ outcome: WebApplePayProbe.Outcome, origin: URL? = nil, generation: Int? = nil) {
+        // A probe that started before a reset measured a state the caller has thrown away.
+        if let generation, generation != self.generation { return }
+        let servedFromCache = persistedReadiness
+        cachedReadiness = outcome.readiness
+        probedOrigin = origin
+        probeInFlight = false
+        // An unknown outcome measured nothing, so the persisted value stays as it is rather
+        // than being replaced by an absent measurement.
+        if !outcome.readiness.isUnknown {
```
salami, line 154: Hmm. I wonder if we actually should still cache for some unknown cases, or maybe even all unknown cases. I lean towards being more sensitive in reducing delays to the on-launch request over timely accuracy of the apple pay probe result

Sources/Helium/HeliumCore/HeliumController.swift
```diff
@@ -39,6 +39,9 @@ class HeliumController {
                     endpoint: fetchedConfig.segmentAnalyticsEndpoint,
                     overrideIfNewConfiguration: true
                 )
+
+                // Needs the fetched Paddle client token to know which origin to probe.
+                WebApplePayAvailability.shared.refreshIfNeeded()
```
salami: I thought the idea was to refresh before we call on launch so that we can actually use it for targeting on new users. I feel like we should await this before the on launch request. We still have the timeout on it. And this would further emphasize the move to a static domain instead of relying on the Paddle token for that.

Sources/Helium/HeliumPaywallPresentation/HeliumPaywallPresenter.swift
salami, line 1: The changes in this file don't seem related to the PR. Why is removeLoadingObservers being moved inside the Task of handleDownloadComplete is that intentional?

Sources/Helium/HeliumPublic.swift
```diff
@@ -832,6 +834,17 @@ public class HeliumConfig {
...
 
+    /// Measures whether the browser external web checkout hands off to can actually pay with
+    /// Apple Pay, and reports it to Helium so a workflow can target on it.
+    ///
+    /// The measurement runs off an offscreen web view at launch, costs up to two seconds of
+    /// background work on the first launch, and is then served from cache and refreshed once
+    /// per launch. Nothing about presentation waits on it.
+    ///
+    /// Defaults to `false`, in which case no measurement runs and every user is reported as
+    /// Apple Pay ready.
```
salami: This will need to be updated. Noting the trade-offs And perhaps that is a new option. I also don't know how explicit we should be in the details. saying "can actually pay with Apple Pay" is false and looks bad. Our current logic is actually fairly accurate. This is optimization.

Sources/Helium/HeliumCore/HeliumController.swift
Arhan13: @salami making a fix for this, Devin, please respond here once its fixed!

Sources/Helium/HeliumPublic.swift
```diff
@@ -67,7 +67,18 @@ public class Helium {
...
+        // measurement straight away and re-measures behind the request.
+        if WebApplePayAvailability.shared.needsMeasurementBeforeLaunch() {
+            Task { @MainActor in
+                await WebApplePayAvailability.shared.refreshAndWait()
+                fetchController.downloadConfig()
+            }
+        } else {
+            WebApplePayAvailability.shared.refreshIfNeeded()
+            fetchController.downloadConfig()
+        }
```
salami: I think this logic would more cleanly live in fetchConfig of Helium since it's already async. This would resolves coderabbit's reset concerns too.

And we could even do it at the same time as `await AppStoreCountryHelper.shared.fetchStoreCountryCode()` for potential parallel performance gains and it keeps prefetching logic together

Sources/Helium/HeliumCore/StripeCheckout/WebApplePayAvailability.swift
salami: But now it does block on-launch. Maybe worth reconsidering?

Sources/Helium/HeliumCore/StripeCheckout/WebApplePayAvailability.swift
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    /// Starts the one refresh this launch gets. Returns immediately; the result lands in the
+    /// cache and in storage when the probe completes.
+    func refreshIfNeeded() {
+        guard let origin = claimProbe() else { return }
+
+        Task { @MainActor in
+            let outcome = await WebApplePayProbe(origin: origin).run()
+            apply(outcome)
```
salami: nit - this could just call refreshAndWait here

Sources/Helium/HeliumCore/StripeCheckout/WebApplePayAvailability.swift
salami: My only fear is if a device continually return unknown on repeated on-launches. Is that possible? Or does the probedOnAPreviousLaunch flag prevent against this? TBH I don't understand the need for both hasProbedKey and persistedReadinessKey persistence.

But regardless I do agree we should not wipe a cached known result with unknown ever.

Sources/Helium/HeliumCore/StripeCheckout/WebApplePayAvailability.swift
```diff
@@ -0,0 +1,201 @@
...
+        guard ApplePayHelper.shared.canMakePayments() else { return .unknown(.deviceCannotPay) }
+        return cachedReadiness
+    }
+
+    /// Readies the value the launch request carries. A device that has never probed has
+    /// nothing to report, so it waits for a measurement, bounded by the probe's own timeout;
+    /// any later launch returns at once and measures behind the request. Waiting is keyed on
+    /// having probed rather than on holding a measurement, so a probe that keeps failing
+    /// costs its timeout once rather than on every launch.
+    func prepareForRequest() async {
```
salami, line 61: Nice I like the consolidation here

salami (approved):

JesusRojass (approved):

salami:
The probe was initiated by a button though? I think most of my delay is because I'm starting the app and calling `initialize` at the same time. No loading UI and then pressing a button. To be fair, I am running it directly from Xcode, so that could contribute to the slowdown too. But my point still stands that the timeout limit needs to be enforced as best as we can. I will do some additional testing with the latest changes.

Arhan13:
> The probe was initiated by a button though? I think most of my delay is because I'm starting the app and calling `initialize` at the same time. No loading UI and then pressing a button. To be fair, I am running it directly from Xcode, so that could contribute to the slowdown too. But my point still stands that the timeout limit needs to be enforced as best as we can. I will do some additional testing with the latest changes.

Let me try this too

Arhan13:
Ran the production probe from this branch on my iPhone 16 Pro, started by `prepareForRequest()` inside `initialize`, no button, cold launches via devicectl.

| Condition | Launch held by `prepareForRequest()` | Probe `durationMs` | Timed out |
|---|---|---|---|
| First-install path, no debugger (8) | median 361 ms, worst 526 | median 272 ms | 0 of 8 |
| Returning-user path, no debugger (8) | median 21 ms | median 261 ms | 0 of 8 |
| First-install path, lldb attached (6) | median 1252 ms, worst 1462 | median 645 ms | 0 of 6 |

So the debugger alone is about 3.5x, and most of the extra is the probe waiting for the main thread while `initialize` runs, which matches what you described. Xcode's Run adds more on top of bare lldb, so I think that explains your numbers.

Agree the cap needs to hold regardless. It wasn't reached in any run, so it's still only verified by reading the code. Happy to force a slow probe and confirm the request leaves at 2 s.

salami:
I confirm my approval.

I did more testing running through Xcode with phone connecting wirelessly. (I deliberately want to stress test the non-happy-path.) With the new changes the longest delay I found before config request was about 3.5 seconds. But more often it was 2500 ms or less.

I also tried DispatchSourceTimer to see if that saves time but did not see any meaningful difference.

My only suggestion is adding an additional field on the existing web_apple_pay_probe_completed observability event to track the full duration of awaitMeasurement. This is what I was looking at, and I think it is a closer measurement of the actual delay before the config request goes out.

PR #361, opened 2026-09-17
HEL-6939 Add an in-app web view web checkout style

salami:
Adds `inAppWebView` as a third in-app browser style, alongside the two Safari ones landed in #353. The server selects it per paywall through the same `webCheckoutBrowserStyle` field; absent or unrecognized still resolves to `externalBrowser`, so nothing changes until the server sends it.

It needs no changes to `ExternalWebCheckoutManager` — the completion path built for Safari generalises unchanged, which is the main thing this PR demonstrates.

## How it differs from the Safari styles

**The redirect never leaves the app.** A custom scheme cannot load in a web view, so it is caught in `decidePolicyFor` and handed to `Helium.handleURL` — the same entry point a host app forwards to — keeping every style on one completion path. Only Helium's own redirects are intercepted; anything else the checkout navigates to is allowed through.

**No loading cover is needed.** The web view paints its own background before the first load, so the wait for the first byte shows the app's colour rather than white. Safari needs a whole cover for this because that region has no API. A spinner appears only if the load takes longer than 250ms, so a fast open shows nothing rather than a flicker.

**Full screen, no chrome.** No navigation bar, dismiss button or host name: the page fills the screen the way the paywall beneath it does, edge to edge with the same scroll insets, and owns its own way out. Because nothing native can close it, a first load that never renders closes the browser and returns the user to the paywall — otherwise a failed load would strand them behind a blank screen with nothing to tap.

**`WKUIDelegate` is implemented** even though nothing reaches it today. Checkout is Apple Pay only and opens no popups, but a web view drops `target="_blank"` and `window.open` when the delegate is absent, with no error at all — the tap simply does nothing. That failure would be close to undiagnosable if the page ever gained a payment method needing one.

## Shared with the Safari styles

Dismissal reporting moves to a `WebCheckoutBrowserViewController` base class, since closing is the only signal that an in-app checkout ended for any of them.

## Risk

This is the style whose page the SDK can read and script, which is what makes it the hardest to defend under the external purchase link rules — more so than the Safari styles, which are themselves already experimental. It stays server-controlled for the same reason: it can be turned off without an app update.

## Testing

Not yet exercised against a real checkout. Outstanding: an Apple Pay purchase end to end, cancel and payment-failure redirects, the failed-load escape, and whether the pay button clears the home indicator now that the page runs edge to edge.

No 3DS testing: checkout is Apple Pay only, which carries its own authentication, and this is US-only where 3DS is not regulatory. So the off-site navigation and popup paths are unreachable in practice rather than merely untested.

---

> [!NOTE]
> **High Risk**
> Changes web checkout presentation and when purchase observations run after browser dismiss/cancel, in payment and entitlement-sensitive paths; the in-app web view style also increases compliance surface versus Safari-hosted checkout.
> 
> **Overview**
> Adds **`inAppWebView`** as a server-selected **`webCheckoutBrowserStyle`**, opening checkout in a full-screen **`WKWebView`** alongside the existing Safari sheet/full-screen and external browser options.
> 
> **Checkout presentation** is refactored around a shared **`WebCheckoutBrowserViewController`** and a **`WebCheckoutBrowserDismissal`** reason (**.closed** vs **.neverLoaded**). **`ExternalWebCheckoutManager`** uses that signal so a browser that never loads drops the checkout observation instead of running a post-dismiss entitlement check; Safari now auto-closes on failed initial load with the same **neverLoaded** path.
> 
> The new **`WebCheckoutWebViewController`** intercepts only Helium redirect URLs in **`decidePolicyFor`** and forwards them to **`Helium.handleURL`**, shows a delayed loading spinner, closes on a failed first load, and implements **`WKUIDelegate`** for **`target="_blank"`** loads.
> 
> **Paywall previews** decode and pass **`webCheckoutBrowserStyle`** from the control panel into preview trigger config (main preview only; second try stays unset).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7f83f7c5a84c48595f503f62c0e455a993726844. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added an in-app web checkout option that opens checkout directly within the app.
  - Checkout pages support navigation, pop-ups, loading feedback, and automatic handling of Helium redirect links.
  - In-app checkout closes automatically when content cannot be loaded.
  - Paywall previews can now apply the selected web checkout browser style.
  - Existing Safari-based checkout options remain available.

- **Bug Fixes**
  - Improved checkout handling when a browser closes normally or fails to load, preventing unnecessary purchase checks after failed loads.


JesusRojass (approved):
Awesome!

